PR #31, opened 2017-02-09
Source files of restructured documentation

david-pl:


bastikr:
Just a copy of the comment I made in another pull request which I think is more relevant here:

I just realized that we also commit the png images that are created in the notebooks. Since this bloats the size of the repository we should really try to avoid it. nbconvert offers an execute function (https://nbconvert.readthedocs.io/en/latest/execute_api.html) so that we can just store the inputs and only calculate the outputs when we rebuild the documentation. This has the additional advantage that we can always be sure that the code is still working with the newest version.

PR #32, opened 2017-02-09
Moved files in subfolder and generated updated docs from new source files

david-pl:


bastikr:
I just realized that we also commit the png images that are created in the notebooks. Since this bloats the size of the repository we should really try to avoid it. nbconvert offers an execute function (https://nbconvert.readthedocs.io/en/latest/execute_api.html) so that we can just store the inputs and only calculate the outputs when we rebuild the documentation. This has the additional advantage that we can always be sure that the code is still working with the newest version.

bastikr:
Also the .nojekyll file has to stay in the top level. It stops the automatic testing to not run on the gh-pages branch.

bastikr:
I just realized that this branch only commits images to the gh-pages branch which is of course necessary. However, we should first do all the planned documentation changes  in the master branch and then just rebuild the documentation once and then commit it. Is this okay for you David?

david-pl:
Sure thing. I will then simply continue to commit the source files to the docs branch of my fork of the repository. However, in order to work on the website I think it is necessary to move the files to the docs sub-folder. Should I convert back to the first commit on this branch and then you merge it?

bastikr:
Yes use the first commit and maybe also move the .buildinfo file. Then just merge it!

PR #37, opened 2017-02-20
Simultaneous diagonalization

david-pl:


src/spectralanalysis.jl
```diff
@@ -158,4 +158,54 @@ Nmax (optional)
...
+    check_eval = isapprox(Ops[i].data*v[:, j] - evals[i][j]*v[:, j], zeros(length(d)); atol=atol, rtol=rtol)
+    check_eval ? nothing : error("Simultaneous diagonalization failed!")
+  end
+
+  index = sortperm(real(evals[1][:]))
+  evals_sorted = [real(evals[i][index]) for i=1:length(Ops)]
+  evals_sorted, v[:, index]
+end
+
+simdiag{T <: SparseOperator}(Ops::Vector{T}; kwargs...) = simdiag([full(A) for A=Ops]; kwargs...)
```
bastikr: Don't do automatic conversions from a sparse to a full operator. Might use too much memory. Throw an error instead with a message that only dense operators can be diagonalized.

src/spectralanalysis.jl
```diff
@@ -158,4 +158,54 @@ Nmax (optional)
...
+
+atol (optional)
+  kwarg of Base.isapprox specifying the tolerance of the approximate check
+  Default is 1e-14.
+
+rtol (optional)
+  kwarg of Base.isapprox specifying the tolerance of the approximate check
+  Default is 1e-14.
+"""
+
```
bastikr: No newlines between docstrings and functions.

src/spectralanalysis.jl
```diff
@@ -158,4 +158,54 @@ Nmax (optional)
...
+    A == dagger(A) ? nothing : error("Non-hermitian operator given!")
+  end
+
+  d, v = eig(sum(Ops).data)
+
+  evals = [Vector{Complex128}(length(d)) for i=1:length(Ops)]
+  for i=1:length(Ops), j=1:length(d)
+    evals[i][j] = (v[:, j]'*Ops[i].data*v[:, j])[1]
+    check_eval = isapprox(Ops[i].data*v[:, j] - evals[i][j]*v[:, j], zeros(length(d)); atol=atol, rtol=rtol)
+    check_eval ? nothing : error("Simultaneous diagonalization failed!")
```
bastikr: Why compare it to a zeros array. Can't we do isapprox(a, b; ...)
And I think it's nicer to do
if !isapprox(...)
   error(...)
end

src/spectralanalysis.jl
```diff
@@ -158,4 +158,54 @@ Nmax (optional)
...
+rtol (optional)
+  kwarg of Base.isapprox specifying the tolerance of the approximate check
+  Default is 1e-14.
+"""
+
+function simdiag{T <: DenseOperator}(Ops::Vector{T}; atol::Real=1e-14, rtol::Real=1e-14)
+
+  # Check input
+  for A=Ops
+    A == dagger(A) ? nothing : error("Non-hermitian operator given!")
```
bastikr: Do we need complete equality or is isapprox good enough?

PR #39, opened 2017-02-21
Faster creation of CompositeBasis in some cases.

bastikr:


david-pl:
This is the same commit as in #40.

PR #77, opened 2017-03-13
Changed the implementation of spectranalysis using Julia's eig function

david-pl:


bastikr:
Sorry I forgot that this pull request was still waiting... Can you change that the imported functions that are overloaded in a module are not reexported (ishermitian, eig, ...)? It's not necessary and is at the moment not done for any other function.

PR #185, opened 2018-01-02
Equate ⊗ with `tensor`

Datseris:
This way they are the same object (as intended I think?) and they share the same documentation string.


Datseris:
I have tested whether this change affects the functions and number of arguments, and it is fine, there were no problems.

Issue #164, opened 2017-07-12
Combined time evolution of quantum, classical, stochastic, etc. systems


david-pl:
A more general interface to pass quantum systems in combination with others to an ODE solver (for example DifferentialEquations.jl) would be nice. This is currently a bit tedious since it requires ugly modifications of modules (see e.g. a semi-classical system where the classical differential equations are stochastic [here](https://github.com/david-pl/QuantumOptics.jl-examples/tree/noisy-cavity-cooling)). It's probably better to this after #92.

ChrisRackauckas:
Can you describe what exactly you're looking for?

david-pl:
We're looking to have arbitrary combinations of quantum systems and classical systems as we already do with the `semiclassical` module, but now also allowing for stochastic terms on either or both (quantum and classical) parts.

So basically, we can add a method for `schroedinger_dynamic` and `master_dynamic`, where when a function `fstoch` is specified, it spawns an `SDEProblem` rather than an `ODEProblem`. This can identically be done for `semiclassical` systems. We might have to make an additional distinction there, though, to specify whether the noise applies to the quantum or classical part (something like `fstoch_quantum`, `fstoch_classical`).

Stochastic Schrödinger and master equations #152 describing e.g. homodyne detection are then a special case, where a specific form of noise affects the quantum part.

ChrisRackauckas:
That sounds good. A very similar `integrate!` front-end needs to then be added to handle stochastic, or it needs to be extended a bit. All that's really required is a quick wrapper to `recast!` to states before calling the user-derivatives and then adding the callback which does saving via `fout`. But it's so similar that it may just want to type-check and add a few things in order to re-use most of the code. Let me know if you want help here.

david-pl:
@ChrisRackauckas I have started implementing stochastic equations in the [stochastic barnch](https://github.com/qojulia/QuantumOptics.jl/blob/stochastic/src/stochastic.jl). So far only the Schrödinger equation is implemented. Before continuing I wanted to ask if you could take a glance at the code and tell me whether I got things right.

To clarify what I did so far:

[...]

Issue #92, opened 2017-03-24
Use DifferentialEquations.jl as ODE solver


bastikr:

ChrisRackauckas:
Hey, what problem are you looking to benchmark/profile? I would be willing to help out, and am looking to profile v0.5 --> v0.6 changes anyways. I know there are a few regressions due to internally changing to broadcast and `@.` removing `muladd`s, but OrdinaryDiffEq.jl should still be in very good shape (and all of the issues I've found have been traced to Base compiler optimizations suddenly missing... not much we can do there but wait...)

Instead of taking on the full dependency, you can use the minimal dependency versions instead. If you make it so that way the user has to pass the algorithm to solve, you can directly depend on DiffEqBase only, which is a really low dep library (in fact, you already depend on it through ODE). If you want to set the algorithm, for example to `DP5()` or `Tsit5()`, you will need to depend on OrdinaryDiffEq.jl.

Let me know what you need and I'd be willing to profile and PR.

bastikr:
Hey Chris, thank you for offering your help! I saw that you already fixed a few things in the profiling directory, but I used this code only while I developed the corresponding functionality and most of it is terrible out of date. All the real benchmarking is currently done in https://github.com/qojulia/QuantumOptics.jl-benchmarks, and the results are displayed on our website (which is not completely up to date since we are doing some restructuring). Most functionality is pretty well covered, only the (stochastic) mcwf solver is missing.

The dependence on the ODE package is also a little bit out of date and was mostly used to test that our own ode solver worked correctly.

It's great to hear that there are versions that don't have as many dependencies as the complete DifferentialEquations.jl package. Transitioning the non-stochastic methods should be pretty much straight forward since the actual ode solver is only called at one single point in the code. The monte carlo solver will probably be a little bit more work but also shouldn't be too hard. My plan for the near future is to release version 0.4 in the next two weeks and immediately afterwards I will start working on this. As soon as I have first benchmark results I will share them here. Of course if anyone else is motivated to work on this please just go ahead - help is always welcome!

ChrisRackauckas:
>the actual ode solver is only called at one single point in the code

Can you point me to that spot?

bastikr:
It's in src/timeevolution_base.jl line 20. (https://github.com/bastikr/QuantumOptics.jl/blob/master/src/timeevolution_base.jl#L20)

ChrisRackauckas:
Using `test_dopri.jl`, I did:

[...]

Issue #189, opened 2018-01-03
Change Ket, Bra etc to be parametrically typed


Datseris:
The structures `Ket` and `Bra` are **type-unstable**. The field `basis` is not strictly typed. This:
```
basis::Basis
```
is ambiguous, since any `Basis` can be part of a `Ket`, meaning that its fields are not type-stable.

The parametrisation must happen on the basis level, e.g. 
```julia
abstract type StateVector{B} where {B<:Basis} end

"""
    Bra(b::Basis[, data])
Bra state defined by coefficients in respect to the basis.
"""
type Bra{B} <: StateVector{B} where {B<:Basis}
    basis::B
    data::Vector{Complex128}
    function Bra(b::Basis, data)
        if length(b) != length(data)
            throw(DimensionMismatch())
        end
        new(b, data)
    end
end
```
In general this may have from very big to very miniscule impact on performance (it has to be measured), However, it hinders multiple dispatch greatly, as one cannot specialize methods on the type of basis (which is required for other changes) and has to use `if` branches with `typeof()` checks.

This will be the first change I will do, since it will allow everything else to be done. It is a pretty huge change though so it will take time to be done properly...

Unfortunately the story with `CompositeBasis` which has `bases::Vector{Basis}` is quite different. The container is abstractly typed which forbids inference, but fixing this problem is quite hard. I had the same issue here: https://github.com/JuliaDynamics/DynamicalBilliards.jl/issues/30 and some answers are here: https://discourse.julialang.org/t/is-there-a-way-to-forward-getindex-for-tuples-in-a-type-stable-way/2889/22

This will be a much later fix though. First type-parametrization is in order.

ChrisRackauckas:
Everything is integrated internally on vectors of complex numbers which is why this won't have much of an effect on performance. However, the save containers which are returned to users are not strictly typed, so the resulting data analysis afterwards is effected and not benchmarked.

Datseris:
I also do not think that the difference in performance will not be massive, however it will allow multiple dispatch on the basis type, which I think is super great for Julian programming.

ExpandingMan:
I'm just sort of an interested observer here and have not dug into this package much, but I'm curious, is there some reason this effort was abandoned?  It does seem that at the very least the ability to dispatch on bases or at least the dimensionality of the Hilbert space as @Datseris suggested would be highly desirable.

david-pl:
[...]

PR #191, opened 2018-01-09
Change internals to Diffeq

ChrisRackauckas:
This sets up the internals to use DiffEq. All of the tests pass. There is a breaking change in that the steady state parts now return the solution array (with saveat matching the giving timespan, so it's the same but includes the initial point). Other than that, nothing changes. This requires DiffEqCallbacks v0.7.0 which should be released shortly.

As for speed, benchmarks show that as setup it's roughly the same on very non-stiff problems. The previous integrator's `reltol=1e-7` takes as much steps as this one's `reltol=1e-6`, and with these the DiffEq one is ever so slightly faster. But true benchmarks should take into account error as well. The QO benchmarks should be run again like in https://github.com/qojulia/QuantumOptics.jl/issues/92 . Note that the current DiffEq infrusturcture takes a hit on very cheap problems due to Julia's keyword argument handling and lack of constant propogation, but those are both fixed in v0.7 so we'll get a free speedup when the next version of Julia is released.

But, the DiffEq DP5 uses PI adaptive time stepping which will be more stable and by far more efficient on semi-stiff equations.  And that's mostly the point. I wouldn't expect anything major on simple non-stiff equations like in the tests, but if you increase the stiffness of the PDE, then DiffEq allows a lot more options to handle stiff equations (whereas explicit Runge-Kutta methods simply fail). However, the option to pass integration algorithms to override the internal default is not part of this PR (but it would be very simple). Also, by changing `ODEProblem` to things like `DDEProblem` or `SDEProblem` this (with different algorithm choices) will solve other types of differential equations as proposed in https://github.com/qojulia/QuantumOptics.jl/issues/164 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
     - osx
     - linux
 julia:
-    - 0.5
     - 0.6
     - nightly
 matrix:
--- a/README.md
+++ b/README.md
@@ -8,8 +8,6 @@ More information, documentation and examples can be found on our website http://
 ### Development status
 
 **Latest release:**
-  * [![Status of latest release on julia 0.4][pkg-0.4-img]][pkg-0.4-url]
-  * [![Status of latest release on julia 0.5][pkg-0.5-img]][pkg-0.5-url]
   * [![Status of latest release on julia 0.6][pkg-0.6-img]][pkg-0.6-url]
 
 **Master branch:**
@@ -54,12 +52,6 @@ Additionally, you can contact us on Gitter: [![Chat on Gitter][gitter-img]][gitt
 
 [codecov-url]: https://codecov.io/gh/qojulia/QuantumOptics.jl
 [codecov-img]: https://codecov.io/gh/qojulia/QuantumOptics.jl/branch/master/graph/badge.svg
-
-[pkg-0.4-url]: http://pkg.julialang.org/?pkg=QuantumOptics&ver=0.4
-[pkg-0.4-img]: http://pkg.julialang.org/badges/QuantumOptics_0.4.svg
-
-[pkg-0.5-url]: http://pkg.julialang.org/?pkg=QuantumOptics&ver=0.5
-[pkg-0.5-img]: http://pkg.julialang.org/badges/QuantumOptics_0.5.svg
 
 [pkg-0.6-url]: http://pkg.julialang.org/?pkg=QuantumOptics&ver=0.6
 [pkg-0.6-img]: http://pkg.julialang.org/badges/QuantumOptics_0.6.svg
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,4 +2,4 @@ julia 0.6
 Compat 0.17
 Roots
 OrdinaryDiffEq 2.34.0
-DiffEqCallbacks 0.7.0
+DiffEqCallbacks 0.7.1
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
-julia 0.5
+julia 0.6
 Compat 0.17
 Roots
 OrdinaryDiffEq 2.34.0
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,5 @@
 julia 0.5
 Compat 0.17
 Roots
+OrdinaryDiffEq 2.34.0
+DiffEqCallbacks 0.7.0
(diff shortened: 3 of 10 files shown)


ChrisRackauckas:
Re-running tests because the release that this needed is now in METADATA. Also, I just noticed that this doesn't do anything special for the single-trajectory Monte Carlo stuff, so I am not sure why tests pass there. Those jumps need to be `DiscreteCallbacks`.

Datseris:
Yeah maybe a good idea to not test on "latest" :P

david-pl:
I ran the benchmarks. They look very promising.

bench-master

And even though, as you said, this is not doing anything special for Monte Carlo approaches, it is also not slowing things down.

bench-mcwf

In the above, the orange line is your PR.

There is a problem though, that I noticed while benchmarking. In the benchmarks, we use an `fout`.
When using this code (similar to the benchmarking code):
```julia
b = FockBasis(9)
a = destroy(b)
at = create(b)
n = number(b)

tspan = [0:1.0:10;]

H = a + at
J = [a]
rates = [1.0]

Ψ₀ = coherentstate(b, 0)

tout, ρt = timeevolution.master(tspan, Ψ₀, H, J; rates=rates)
exp_n = real.(expect(n, ρt))
```
the amount of points stored correspond to `length(tspan)`. However, using an `fout` like so
```julia
exp_n = Float64[]
fout(t, ρ) = push!(exp_n, real(expect(n, ρ)))
timeevolution.master(tspan, Ψ₀, H, J; rates=rates, fout=fout)
```
results in an `exp_n` of length `length(tspan)+1`. As it turns out, the first save is pushed twice to `exp_n`, which shouldn't be there of course. Also, the call to `master` returns a non-void object even if the `fout` is defined, whereas previously it returned `nothing`.

The build checks keep failing, because we still have support for Julia v0.5, for which your latest Callbacks is not available. We should probably drop support for v0.5 anyway.

david-pl:
The call to `fout` when specifying the type for `SavedValues` in this line
```julia
out = DiffEqCallbacks.SavedValues(Float64,typeof(fout(tspan[1], state)))
```
in combination with the in-place `push!` used in the above example is the cause of the problem I think.

ChrisRackauckas:
>The build checks keep failing, because we still have support for Julia v0.5, for which your latest Callbacks is not available. We should probably drop support for v0.5 anyway.

Support for v0.5 is locked in METADATA. No packages can release new versions with v0.5. I'll go ahead and upgrade the build scripts.

ChrisRackauckas:
>in combination with the in-place push! used in the above example is the cause of the problem I think.

Oh, I forgot to mention that in the new version there is a breaking change that you no longer have to build arrays to `push!` into, you just return the values in `fout` and they will be the solution values given by the function. I could change this but it seemed much more intuitive IMO. See the way the tests are setup.

david-pl:
Oh, I could have caught that...
Hm, I agree that, e.g. like in one of the test,
```julia
function fout(t, rho)
    expect(op1, rho)
end
```
is more intuitive than using `push!`.

Properly inferring the type of `fout` (as you say in the TODO comment) would avoid this breaking though right?

ChrisRackauckas:
>Properly inferring the type of fout (as you say in the TODO comment) would avoid this breaking though right?

Yes, it would be the same to the user, but I would just use `Base.core.inference` calls to know the output type instead of actually computing it once. Let me go ahead and add that right now. I didn't want to do it before I got everything working because it's a little bit tedious, but now that everything is together it makes sense to handle it.

ChrisRackauckas:
I edited the inference to work without the `fout` call. Also users can pass an algorithm choice. It's an optional arg for type-stability, but it can be made a kwarg. That won't be stable of course until Julia v0.7, but if it's breaking it might as well just happen now?

david-pl:
I don't think the additional argument can break much. On the other hand, I don't know how much this instability will even affect things. So I think either way is fine.

ChrisRackauckas:
MCWF is fixed to now use the DiffEq callbacks. `alg` is a new keyword argument for algorithms. Because of https://github.com/qojulia/QuantumOptics.jl/issues/189, the scope it's in is already type-unstable and probably won't be fixed before v0.7 so it might as well be in its "final form"

ChrisRackauckas:
I think that's everything. DDEs/SDEs/etc. can be a separate PR. But this should be everything switched over to DiffEq internals, with the breaking changes that `fout` now doesn't require user `push!`ing and instead saves the return, along with the SS saving at `tstops` which in the tests includes the first point.

david-pl:
I ran the benchmarks on MCWF again. Looks like your final changes will give us quite a performance boost.

bench-mcwf

The green one here is your latest commit, the others are the same as before.

So let's merge this beauty!

Issue #190, opened 2018-01-05
Implement `gemm!` for (LazyTensor, LazyProduct) combo


Datseris:
For reference:
```julia
xmin = -30
xmax = 30
b_position = PositionBasis(xmin, xmax, 100)
b_momentum = MomentumBasis(b_position)

b_positiony = PositionBasis(xmin, xmax, 100)
b_momentumy = MomentumBasis(b_positiony)

Txp = transform(b_position, b_momentum)
Tpx = transform(b_momentum, b_position)
Typ = transform(b_positiony, b_momentumy)
Tpy = transform(b_momentumy, b_positiony)

Hkinx = LazyProduct(Txp, momentum(b_momentum)^2/2, Tpx)
Hkiny = LazyProduct(Typ, momentum(b_momentumy)^2/2, Tpy)

b_comp = b_position ⊗ b_positiony

Hkinx_comp = LazyTensor(b_comp, [1, 2], [Hkinx, one(b_positiony)])
Hkiny_comp = LazyTensor(b_comp, [1, 2], [one(b_position), Hkiny])

H = LazySum(Hkinx_comp, Hkiny_comp)
```

Now doing
```julia
ψx = gaussianstate(b_position, -10.0, 1.0, 2)
ψy = gaussianstate(b_positiony, 0, 0.5, 2)
ψ = ψx ⊗ ψy

T = collect(linspace(0., 10.0, 50))
tout, C = timeevolution.schroedinger(T, ψ, H)
```
gives the error that `gemm!` is not defined for LazyTensor and LazyProduct. Instead one has to work-around by passing `Hfull = sparse(full(H))` instead of `H` to `timeevolution`.

david-pl:
As a remark, the issue you are having can actually be traced back to the fact that no form of tensor product (lazy or normal) is implemented for the `FFTOperator` type.

To explain: While it is true that `gemm!` is not implement for a `LazyTensor` of `LazyProduct`, it is always possible to rewrite such an instance to a `LazyProduct` of `LazyTensor` which `gemm!` can handle.
[...]

PR #192, opened 2018-01-19
Generalization of FFTOperator to composite bases

david-pl:
This should implement all functionality required to solve #190.

Some more comments:
- Benchmarks need to be run once more.
- This is not a lazy implementation, but using `ket_only` makes it as efficient as saving a `Ket`. Going below this is only relevant if one is neglecting correlations.
- Things might not be entirely type stable here, so there might be some clean-up work to do.
- Using an `FFTOperator` in a `LazyTensor` will still throw the error as in #190. However, this can be avoided by rewriting things to a `LazyProduct`. We should maybe consider making this rewriting automatic or at least provide a more helpful error message.

EDIT:
Also, more rigorous tests are necessary.


david-pl:
I made the index that specifies along which dimensions the FFT is being calculated an optional argument. This is needed in case someone has a composite system with more than one position/momentum basis, but only wants to use the FFT method in a specified subset of these bases.

Also, I ran the benchmarks in 1D again. They still give me different results on different runs and I am not sure why. Still, they look to be quite fine.
bench-fft
The blue line is the new implementation here.

PR #196, opened 2018-01-27
DifferentialEquations v4

Datseris:
[x] master
[x] mcwf
[x] pumped
[x] schroedinger
[x] twolevel

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,5 @@
 julia 0.6
 Compat 0.17
 Roots
-OrdinaryDiffEq 2.34.0
-DiffEqCallbacks 0.7.1
+OrdinaryDiffEq 3.0.1
+DiffEqCallbacks 1.0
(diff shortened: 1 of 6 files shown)


Datseris:
@david-pl what's up with the documentation string of `integrate` ?

```julia
"""
df(t, state::T, dstate::T)
"""
function integrate{T}(tspan::Vector{Float64}, df::Function, x0::Vector{Complex128},
            state::T, dstate::T, fout::Function;
            alg::OrdinaryDiffEq.OrdinaryDiffEqAlgorithm = OrdinaryDiffEq.DP5(),
            steady_state = false, tol = 1e-3, save_everystep = false,
            callback = nothing, kwargs...)

```
?...

Datseris:
Hoorayyyyyy!!! all tests pass locally!

david-pl:
You missed one line in semiclassical.master. All tests should pass now. Other than that, great work. Thanks!

Datseris:
Good, thanks for the help! I had to admit it wasn't an easy process to find all lines, due to the huge amount of files!

david-pl:
We could update to the new DiffEq syntax with much fewer changes by simply rewriting `df_` in `integrate`. This would keep the remaining syntax independent of DiffEqs syntax. Future syntax changes of DiffEq are then easier to handle. This is done in #197.

@Datseris Is there a specific reason for doing it this way?

PR #197, opened 2018-01-27
Update to new DiffEq syntax

david-pl:


Datseris:
OH man... I feel so sad and so devastated.

Because I do not really know the inner workings of QuantumOptics, I was simply running tests and changing the functions that failed one by one... But apparently there was a much smarter way!

Great!

PR #207, opened 2018-03-09
Address Issue "Odd convention for the tensor product? "

mattm1:
I inverse the order of the subsystems, to get the standard output for tensor products. It's an easy fix for the problem.

Because rounding errors clutter the output for large matrices I round them to zero.

The data representation is unchanged with .data. It just changes the output for bra/ket and sparse/dense operators.

### DIFF
--- a/src/QuantumOptics.jl
+++ b/src/QuantumOptics.jl
@@ -102,6 +102,7 @@ using .timeevolution
 using .metrics
 using .spectralanalysis
 using .timecorrelations
+using .printing
 
 
 end # module
--- a/src/printing.jl
+++ b/src/printing.jl
@@ -120,7 +120,7 @@ show(stream::IO, x::Operator) = showoperatorheader(stream, x)
 function showquantumstatebody(stream::IO, x::Union{Ket,Bra})
     #the permutation is used to invert the order A x B = B.data x A.data to A.data x B.data
     perm = collect(length(basis(x).shape):-1:1)
-    if length(perm) == 1 || !_std_order
+    if length(perm) == 1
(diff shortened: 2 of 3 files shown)


david-pl:
Hey. Thanks for your submission!

Even though simply changing the output is a nice idea, I think that this will ultimately lead to more confusion. As one is usually working with the `.data` fields when indexing, but viewing the output in the REPL, this may be misleading. Instead, I think we should simply stick to the current ordering and clearly state it in the docs (explaining that it's due to performance).

As for excluding the rounding errors, this could be nice. I'm just not sure whether you would want to know if your system is making rounding errors when looking at the REPL output...

mattm1:
Hey,
if you are not sure if the user wants to exclude rounding errors, or wants to get the standard ordering of subsystems, then I propose a function
`set_qobj_reploptions(standard_order::Bool = False, rounding_tol = 0.0)`
the default will coincide with the .data output, but the user can change the output at will.

david-pl:
That sounds like a great idea, thanks.

david-pl:
@mattm1 I drafted a function as you suggested. Is this what you had in mind?

Now you can do
```julia
b = SpinBasis(1//2)
x = spindown(b) ⊗ spinup(b)

Ket(dim=4)
  basis: [Spin(1/2) ⊗ Spin(1/2)]
 0.0+0.0im
 1.0+0.0im
 0.0+0.0im
 0.0+0.0im
```
and
```julia
QuantumOptics.set_printing(standard_order=true)
x

Ket(dim=4)
  basis: [Spin(1/2) ⊗ Spin(1/2)]
 0.0+0.0im
 0.0+0.0im
 1.0+0.0im
 0.0+0.0im
```

PR #228, opened 2018-07-25
fix ishermitian for dense and sparse operator

goropikari:
If an operator is Hermitian, both left and right bases should be same.


david-pl:
Good point, thanks for the fix.

PR #231, opened 2018-07-26
Change steady state sorting

david-pl:
@vtorggler is this okay now or are further changes required?


vtorggler:
Maybe it would be better to sort it with respect to absolute value only. Then those eigenvalues with zero real and imaginary part are on top. But I guess it's ok as it is.

But, we also have to remove the normalization (division by trace(ops)) in liouvillianspectrum and add it in the eigenstate function.

david-pl:
@vtorggler Good point. How about now?

vtorggler:
Ok one last annoyance.

In Line 112 I would keep checking if the real part is 0, because I think if there are degeneracies it could be that there is an imaginary (oscillating) part. Rather warn earlier.

And the error message in Line 109 should be adapted. "No zero eigenvalue found." or so.

I should do this :)

david-pl:
Ok then, how about now? ^^

vtorggler:
Perfetto :+1:

PR #235, opened 2018-09-14
Define ' on Operator as dagger()

klickverbot:
This is useful and makes sense mathematically, especially
now that the meaning of `'` has been clarified as `adjoint`.


david-pl:
Good point. Thanks!

PR #246, opened 2019-03-20
before embedding check that the destination basis matches the operator basis

### LINKED ISSUE #240: embed with false indexing overwrites sub-basis
This bit of code
```julia
using QuantumOptics
bf = FockBasis(5)
bs = SpinBasis(1//2)
σ = sigmam(bs)

embed(bf⊗bs,1,σ)
```
results in
```julia
SparseOperator(dim=4x4)
  basis: [Spin(1/2) ⊗ Spin(1/2)]
  [2, 1]  =  1.0+0.0im
  [4, 3]  =  1.0+0.0im
```
which clearly overwrites the `FockBasis`. This should throw an error.

atombear:
closes #240

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -9,7 +9,6 @@ import LinearAlgebra: tr, ishermitian
 import SparseArrays: sparse
 import ..bases: basis, tensor, ptrace, permutesystems,
             samebases, check_samebases, multiplicable
-# import ..operators_sparse: SparseOperator
 import ..states: dagger, normalize, normalize!
 
 using ..sortedindices, ..bases, ..states
@@ -92,7 +91,10 @@ tensor(operators::AbstractOperator...) = reduce(tensor, operators)
 
 
 """
-docstring
+    check_indices(indices::Array)
+
+Determine whether a collection of indices, written as a list of (integers or lists of integers) is unique.
+This assures that the embedded operators are in non-overlapping subspaces.
 """
 function check_indices(indices::Array)
     status = true
@@ -124,6 +126,7 @@ function embed(basis_l::CompositeBasis, basis_r::CompositeBasis,
     @assert length(basis_r.bases) == N
     @assert length(indices) == length(operators)
 
+    # Embed all single-subspace operators.
     idxop_sb = [x for x in zip(indices, operators) if typeof(x[1]) <: Int64]
     indices_sb = [x[1] for x in idxop_sb]
     ops_sb = [x[2] for x in idxop_sb]
@@ -135,6 +138,7 @@ function embed(basis_l::CompositeBasis, basis_r::CompositeBasis,
 
     embed_op = tensor([i ∈ indices_sb ? ops_sb[indexin(i, indices_sb)[1]] : identityoperator(T, basis_l.bases[i], basis_r.bases[i]) for i=1:N]...)
 
+    # Embed all joint-subspace operators.
     idxop_comp = [x for x in zip(indices, operators) if typeof(x[1]) <: Array]
 
     for (idxs, op) in idxop_comp
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -6,8 +6,10 @@ export AbstractOperator, DataOperator, length, basis, dagger, ishermitian, tenso
 
 import Base: ==, +, -, *, /, ^, length, one, exp, conj, conj!, transpose
 import LinearAlgebra: tr, ishermitian
+import SparseArrays: sparse
 import ..bases: basis, tensor, ptrace, permutesystems,
             samebases, check_samebases, multiplicable
+# import ..operators_sparse: SparseOperator
 import ..states: dagger, normalize, normalize!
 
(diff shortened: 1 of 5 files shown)


david-pl:
I changed the syntax as discussed in #240. I'm sure things could be more elegant, but at least everything seems to work properly like this. Let me know if you have any suggestions before I merge this.

david-pl:
I didn't go through your code in detail, so correct me if I'm wrong, but it seems like your function handles the case of a single (composite) operator being embedded in a product basis only. This case is covered by the method I wrote as well. So yes, I think this is a bit redundant.

atombear:
Yes, my method handles the case of one operator being embedded in any subspace of a composite hilbert space. so, for 8 qubits, D=2^8, and a CNOT will be embedded appropriately in any 4d supspace, meaning in any 2 indices. I think your method does not distinguish between an operator acting an different, but equivalent, sub-bases. for example, for a 3x5 operator being embedded in a space that is 3x3x5, your method gives the same output whether the indices are [1, 3] or [2, 3], but those should yield different operators.

atombear:
so i think my position is as follows: the original code `(indices, ops)` tried to match every `op` to an `index`, but had the problem that the matching would not bother checking that the `op` matched the basis corresponding to the `index` in which it was being inserted.

the new method `(indices, ops)` tries to do two things:
- assert that the size of the basis matches, thereby fixing the bug
- try to account for composite spaces by index counting

however, there is an oversight in that it assumes operators in composite spaces correspond to consecutive basis indices. i think it is impossible to correct for this without the index wrangling. Additionally, it assumes the following syntax: `[1,2,3], [op12, op3]` to place a composite operator `op12` in the space `[1,2]`, and a third operator `op3` in space `3`. perhaps the syntax `[[1,2],3], [op12, op3]` would be clearer?

the new `(indices, op)` method generically embeds an operator in a composite hilbert space in an even larger hilbert space, by doing the appropriate index wrangling.

i think it is true that the more general and bug-free `(indices, ops)` method can be written in terms of the new `(indices, op)` method, allowing for non-consecutive indices. my proposal is to do this, and i am glad to take on the task if that makes sense to you. additionally, i am happy with any syntax, though the second i think is clearer, eg: `[[1,3],2], [op13, op2]`.

atombear:
i think i have pushed a test-passing working version of what i described. there are a few more tests i'd like to add, but i think the code is at least correct (finally). a review would be much appreciated at this point.

david-pl:
I appreciate all your work. The changes look good and I agree with the new syntax, which makes a lot more sense especially in the case that I oversaw (embedding composite operators on bases such as e.g. [1,3]). I'm also not sure why the tests are failing. Hopefully it was just a hick-up, so I triggered them by closing/re-opening.

src/operators.jl
```diff
@@ -88,22 +89,128 @@ tensor(op::AbstractOperator) = op
...
 
 
+
+"""
+    check_indices(indices::Array)
+
+Determine whether a collection of indices, written as a list of (integers or lists of integers) is unique.
+This assures that the embedded operators are in non-overlapping subspaces.
+"""
+function check_indices(indices::Array)
```
david-pl: It would make sense to move this function into the `sortedindices` module

src/operators.jl
```diff
@@ -88,22 +89,128 @@ tensor(op::AbstractOperator) = op
...
     @assert length(basis_r.bases) == N
     @assert length(indices) == length(operators)
+
+    # Embed all single-subspace operators.
+    idxop_sb = [x for x in zip(indices, operators) if typeof(x[1]) <: Int64]
+    indices_sb = [x[1] for x in idxop_sb]
+    ops_sb = [x[2] for x in idxop_sb]
+
+    for (idxsb, opsb) in zip(indices_sb, ops_sb)
+        @assert (opsb.basis_l == basis_l.bases[idxsb]) || throw(bases.IncompatibleBases())
```
david-pl: The `@assert` statement here (and in the next line) is unnecessary

src/operators.jl
```diff
@@ -88,22 +89,128 @@ tensor(op::AbstractOperator) = op
...
+    embed(basis1[, basis2], indices::Vector, operators::Vector)
+
+Embed operator acting on a joint Hilbert space where missing indices are filled up with identity operators.
+"""
+function embed(basis_l::CompositeBasis, basis_r::CompositeBasis,
+               indices::Vector{Int}, op::T) where T<:AbstractOperator
+    N = length(basis_l.bases)
+    @assert length(basis_r.bases) == N
+
+    @assert reduce(tensor, basis_l.bases[indices]) == op.basis_l || throw(bases.IncompatibleBases())
```
david-pl: Again, `@assert` is not needed here.

src/operators.jl
atombear: great

src/operators.jl
atombear: done

src/operators.jl
atombear: done.

PR #250, opened 2019-05-27
Adding Bloch-Redfield master equation

sd109:
I've added the bloch_redfield_master file which contains a function for calculating the Bloch-Redfield tensor (Liouvillian) and the corresponding time dynamics. I didn't fully understand how the ODE solving was implemented in timeevolution_base so I have just written my own simple ODE solving function (*bloch_redfield_dynamics*) but maybe it would be better to keep this consistent with the rest of QO.jl?


david-pl:
In principle the `integrate` function in `timeevolution_base` just takes the function that is later passed to `DiffEq` (in your case `ρ_dot`) and creates an in-place `ODEProblem`. Rewriting your code to do this should be rather straightforward and I can do it.

Since I'm not so familiar with the Bloch-Redfield formalism: do you have an example, where the result is known, so we can test the implementation?

sd109:
I have checked that the Julia function gives the same numerical results for the Bloch-Redfield tensor and the dynamics as the QuTiP version. So if by "known" you mean that it has been calculated numerically elsewhere then I guess [this example](http://qutip.org/docs/latest/guide/dynamics/dynamics-bloch-redfield.html#equation-qubit) from the QuTiP documentation is something that I can easily recreate as a test (or any of the other tests from the QuTiP github I guess).

If you would prefer a test case that can be checked against an analytical calculation then I would have to think about this some more as I don't know of any of the top of my head but it should be possible to come up with one in principle.

david-pl:
Thanks for pointing out the example. Actually, we don't have to recreate the entire thing, but can just check against the numerical values of the tensor and states after some dynamics.

sd109:
I've still not got round to learning about testing etc. in Julia so don't know how to add the tests myself but here is the Julia version of the QuTiP example which reproduces the same numerical values for the tensor:

```julia
using QuantumOptics

Δ = 0.2 * 2*π
ϵ0 = 1.0 * 2*π
γ1 = 0.5

b = SpinBasis(1//2)
sx = sigmax(b)
sz = sigmaz(b)

H = -Δ/2.0 * sx - ϵ0/2.0 * sz

function ohmic_spectrum(ω)
    ω == 0.0 ? (return γ1) : (return γ1/2 * (ω / (2*π)) * (ω > 0.0))
end

R, ekets = timeevolution.bloch_redfield_tensor(H, [[sx, ohmic_spectrum]])

dense(R).data
```

david-pl:
Alright, I changed the function so it now uses the internal integration routine. There is still some stuff to do. For one, the saving seems to be type-unstable in the current implementation. Also, the back-transformation from the eigenspace of `H` can be more efficient. I will look into that.
We should also include some proper doc-strings as documentation.

Could you check if the changes I made didn't break anything? Are the dynamics the same as before? Note that I also renamed the function for the dynamics to more closely resemble the other routines in the `timeevolution` module.

sd109:
I've checked that your changes didn't break anything and that the tensor and dynamics are the same as before. I've also added docstrings for the tensor and dynamics functions and included a reference to the QuTiP docs since this implementation is based on theirs.

david-pl:
Ok great, thanks! The tests will fail, but that's unrelated and already fixed on master.
I'll see when I can get around to improving the saving and then we're ready to go here.

david-pl:
Could you have one last look at this and let me know if everything's okay from your side? I'll merge then.

sd109:
All looks good to me now, thanks for your help!

PR #251, opened 2019-06-04
created abstractions for quantum information on qubits

atombear:
This is a WIP: I will add (more) tests in the next few days!

### DIFF
--- a/src/pauli.jl
+++ b/src/pauli.jl
@@ -4,6 +4,7 @@ export PauliBasis, PauliTransferMatrix, DensePauliTransferMatrix,
         ChiMatrix, DenseChiMatrix
 
 import Base: ==
+import Base: *
 
 using ..bases, ..spin, ..superoperators
 using ..operators: identityoperator, AbstractOperator
@@ -23,10 +24,9 @@ mutable struct PauliBasis{B<:Tuple{Vararg{Basis}}} <: Basis
     shape::Vector{Int}
     bases::B
     function PauliBasis(num_qubits::Int)
-        type = Tuple{(SpinBasis{1//2} for _ in 1:num_qubits)...}
         shape = [2 for _ in 1:num_qubits]
         bases = Tuple(SpinBasis(1//2) for _ in 1:num_qubits)
-        return new{type}(shape, bases)
+        return new{typeof(bases)}(shape, bases)
     end
 end
 ==(pb1::PauliBasis, pb2::PauliBasis) = length(pb1.bases) == length(pb2.bases)
@@ -61,6 +61,11 @@ end
 
 PauliTransferMatrix(ptm::DensePauliTransferMatrix{B, B, Matrix{Float64}}) where B <: Tuple{PauliBasis, PauliBasis} = ptm
 
+function *(ptm0::DensePauliTransferMatrix{B, B, Matrix{Float64}},
+           ptm1::DensePauliTransferMatrix{B, B, Matrix{Float64}}) where B <: Tuple{PauliBasis, PauliBasis}
+    return DensePauliTransferMatrix(ptm0.basis_l, ptm1.basis_r, ptm0.data*ptm1.data)
+end
+
 """
     Base class for χ (process) matrix classes.
 """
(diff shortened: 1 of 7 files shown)


src/pauli.jl
```diff
@@ -0,0 +1,229 @@
...
+        new{BL, BR, T}(basis_l, basis_r, data)
+    end
+end
+
+DenseChiMatrix(chi_matrix::DenseChiMatrix{B, B, Array{Complex{Float64}, 2}}) where B <: Tuple{PauliBasis, PauliBasis} = chi_matrix
+
+"""
+    pauli_operators(num_qubits::Int64)
+
+Generate a matrix of basis vectors in the Pauli representation given a number
```
ampolloreno: Just learning about Julia's type system - is this actually a matrix?

It seems like matrix is Array{T,2} where T, while this seems to produce something like: Array{Array{Int64,1},1}.
Is an array of arrays formally equivalent to a matrix in julia?

src/pauli.jl
```diff
@@ -0,0 +1,229 @@
...
+    SuperOperator(chi_matrix::DenseChiMatrix)
+
+Convert a χ matrix to its representation as a superoperator.
+"""
+function SuperOperator(chi_matrix::DenseChiMatrix{B, B, Array{Complex{Float64}, 2}}) where B <: Tuple{PauliBasis, PauliBasis}
+    return SuperOperator(PauliTransferMatrix(chi_matrix))
+end
+
+"""
+    ChiMatrix(ptm::PauliTransferMatrix)
```
ampolloreno: Should this also say Dense?

src/pauli.jl
```diff
@@ -0,0 +1,229 @@
...
+==(pb1::PauliBasis, pb2::PauliBasis) = length(pb1.bases) == length(pb2.bases)
+
+"""
+Base class for Pauli transfer matrix classes.
+"""
+abstract type PauliTransferMatrix{B1<:Tuple{PauliBasis, PauliBasis}, B2<:Tuple{PauliBasis, PauliBasis}} end
+
+
+"""
+    DensePauliTransferMatrix(b, b, data)
```
ampolloreno: Should the variable names in the docstring match?

src/pauli.jl
```diff
@@ -0,0 +1,229 @@
+module pauli
+
+export PauliBasis, PauliTransferMatrix, ChiMatrix
+
+import Base: ==
```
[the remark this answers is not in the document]
atombear: in julia it is necessary to import base methods in this way in order to extend them.

src/pauli.jl
```diff
@@ -0,0 +1,229 @@
...
+using ..superoperators: SuperOperator
+using ..operators_dense: DenseOperator
+using ..spin: sigmax, sigmay, sigmaz
+using SparseArrays
+using LinearAlgebra: tr
+
+"""
+    PauliBasis(num_qubits::Int)
+
+Basis for an N-qubit space where `num_qubits` specifices the number of qubits.
```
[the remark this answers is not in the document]
atombear: spiceifices

src/pauli.jl
```diff
@@ -0,0 +1,229 @@
...
+    sop_dim = 2 ^ (2 * num_qubits)
+    data = Array{Complex{Float64}}(undef, (sop_dim, sop_dim))
+    data .= pbv * ptm.data * pbv' / √sop_dim
+    return DenseSuperOperator(ptm.basis_l, ptm.basis_r, data)
+end
+
+"""
+    PauliTransferMatrix(unitary::DenseOperator)
+
+Convert an operator, presumably a unitary operator, to its representation as a
```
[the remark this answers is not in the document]
atombear: i actually do need guidance on this - does it make any sense to perform this operation on non-unitary operators?

src/pauli.jl
atombear: no, as you say, they are different. i am using matrix colloquially.

src/pauli.jl
atombear: yes. i also wonder if the constructors themselves shouldn't be called Dense...

src/pauli.jl
```diff
@@ -0,0 +1,229 @@
...
+"""
+    PauliBasis(num_qubits::Int)
+
+Basis for an N-qubit space where `num_qubits` specifies the number of qubits.
+The dimension of the basis is 2²ᴺ.
+"""
+mutable struct PauliBasis{B<:Tuple{Vararg{Basis}}} <: Basis
+    shape::Vector{Int}
+    bases::B
+    function PauliBasis(num_qubits::Int64)
```
david-pl: Use `::Int` here (and at other points in the file) in order for things to work properly on Win32

atombear: great

atombear:
@david-pl i think this is ready for review!

src/pauli.jl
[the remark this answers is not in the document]
david-pl: I'm not an expert on this, but it seems to me that the PTM representation can be used on any quantum map, i.e. the only condition the operator here has to fulfill is that, when applied to a density operator, we obtain another density operator. In that sense it doesn't have to be a unitary. Or am I mistaken here?

src/spin.jl
```diff
@@ -34,6 +34,18 @@ SpinBasis(spinnumber::Int) = SpinBasis(convert(Rational{Int}, spinnumber))
 
 ==(b1::SpinBasis, b2::SpinBasis) = b1.spinnumber==b2.spinnumber
 
+"""
+    identityoperator(b::SpinBasis)
+
+Pauli ``I`` operator for the given Spin basis.
+"""
+function identityoperator(b::SpinBasis)
```
david-pl: Why is this function necessary? The generic `identityoperator` method works on `SpinBasis` without it

src/pauli.jl
```diff
@@ -0,0 +1,231 @@
...
+The dimension of the basis is 2²ᴺ.
+"""
+mutable struct PauliBasis{B<:Tuple{Vararg{Basis}}} <: Basis
+    shape::Vector{Int}
+    bases::B
+    function PauliBasis(num_qubits::Int)
+        type = Tuple{(SpinBasis{1//2} for _ in 1:num_qubits)...}
+        shape = [2 for _ in 1:num_qubits]
+        bases = Tuple(SpinBasis(1//2) for _ in 1:num_qubits)
+        return new{type}(shape, bases)
```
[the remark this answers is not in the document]
atombear: awesome

src/spin.jl
atombear: i think i realized that after writing this function, as i don't seem to use it...

src/pauli.jl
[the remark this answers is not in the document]
atombear: i think you are right. the code will work certainly for any operator, but i will change the wording if you think that would be more clear.

atombear:
I had forgotten to add composition - for PTM it is just matrix multiplication. for chi, it is not... I have working functionality, though I assume it is far from optimal.

atombear:
@david-pl i believe i have addressed all concerns.

src/pauli.jl
```diff
@@ -0,0 +1,309 @@
...
+        bases = Tuple(SpinBasis(1//2) for _ in 1:num_qubits)
+        return new{typeof(bases)}(shape, bases)
+    end
+end
+==(pb1::PauliBasis, pb2::PauliBasis) = length(pb1.bases) == length(pb2.bases)
+
+"""
+    Base class for Pauli transfer matrix classes.
+"""
+abstract type PauliTransferMatrix{B1<:Tuple{PauliBasis, PauliBasis}, B2<:Tuple{PauliBasis, PauliBasis}} end
```
david-pl: Actually, should this (and also `ChiMatrix`) be subtyped to `SuperOperator`?

src/pauli.jl
```diff
@@ -0,0 +1,309 @@
...
+    ChiMatrix(ptm::DensePauliTransferMatrix)
+
+Convert a Pauli transfer matrix to its representation as a χ matrix.
+"""
+function ChiMatrix(ptm::DensePauliTransferMatrix{B, B, Matrix{Float64}}) where B <: Tuple{PauliBasis, PauliBasis}
+    return ChiMatrix(SuperOperator(ptm))
+end
+
+"""Equality for all varieties of superoperators."""
+==(sop1::Union{DensePauliTransferMatrix, DenseSuperOperator, DenseChiMatrix},
```
david-pl: The `==` method here overwrites the existing one for `DenseSuperOperator`. Also, I would really not use `isapprox` for `==` methods. Rather, define a corresponding method for `isapprox` to do this. The type-checking can also be done via methods, i.e.
```julia
==(sop1::T, sop2::T) where T<:Union{DensePauliTransferMatrix, DenseSuperOperator, DenseChiMatrix} = sop1.data == sop2.data
==(sop1::Union{DensePauliTransferMatrix, DenseSuperOperator, DenseChiMatrix}, sop2::Union{DensePauliTransferMatrix, DenseSuperOperator, DenseChiMatrix}) = false
```
should do the trick. Note, that if you choose to subtype ptms and chi-matrices to `SuperOperator`, the above becomes obsolete and it uses the superoperator method https://github.com/qojulia/QuantumOptics.jl/blob/b46911d2072d9478dcaf47637bc9bda8e4a76311/src/superoperators.jl#L99
Of course, a method for `isapprox` is still needed if you want to have that.

src/pauli.jl
atombear: i don't think so - they are all related by basis transformations, and are thus on equal footing

src/pauli.jl
atombear: done

atombear:
@david-pl i think i have addressed all concerns.

david-pl:
@atombear Thanks for all the work on this!

PR #256, opened 2019-09-20
use ARPACKException from Arpack.jl

KristofferC:
ARPACKException is no longer defined in LinearAlgebra but instead in Arpack.jl (https://github.com/JuliaLinearAlgebra/Arpack.jl/pull/65).

This fixes the tests in the package on 1.3


david-pl:
Thanks for the quick patch!

PR #262, opened 2019-11-26
Type-stable bloch_redfield_tensor

sd109:
Added a couple of type annotations to the bloch_redfield tensor function and also forced Lindblad part of the tensor to be a sparse operator as this was causing the return value to lack type-stablity.

This leads to a noticeable improvement in time and memory allocations used for this function, especially for larger systems and cases with use_secular=false. Happy to provide example code to demonstrate speed-up if required.


david-pl:
Looks good to me, thanks! The test failure was unrelated. I added a quick-fix for it. Will merge as soon as tests pass.

PR #275, opened 2020-08-19
initial attempt at adding unitary propagation

alastair-marshall:
My first attempt at implementing a unitary propagation function in schroedinger.jl

### DIFF
--- a/src/schroedinger.jl
+++ b/src/schroedinger.jl
@@ -1,11 +1,11 @@
 """
     timeevolution.schroedinger(tspan, psi0, H; fout)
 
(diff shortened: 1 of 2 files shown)


alastair-marshall:
I haven't written any tests for it yet though

david-pl:
Some tests would be good. Also, maybe mention that you can compute the propagator in the docstring. Otherwise it looks good, thanks!

alastair-marshall:
Added tests, which pass, for a 2 level system being driven for one Rabi cycle. Result is, excluding global phase, the identity as we expect

PR #307, opened 2021-06-02
Simpler bloch redfield tensor

### LINKED ISSUE #270: bloch_redfield_master.jl
You can see also other more clear implementation 
https://github.com/whaley-group-berkeley/QSpectra/blob/master/qspectra/dynamics/redfield.py
Which can use einsum 
https://nextjournal.com/under-Peter/julia-summer-of-einsum
But here j(w) not double spectral density as qutip use
Also u can do a class for spectral density as done in Qspctra package
(Only support  drude spectral density with Lorentz cut off)
In pyrho package there are many form of spectral density are found in Hamiltonian class
https://github.com/berkelbach-group/pyrho?files=1
UnFortunately i am still beginner to julia but I hope soon can add support

sd109:
Significantly improves both efficiency and readability of Bloch-Redfield tensor implementation. Also utilizes multi-threading when julia session is running with more than one thread available. I'd say this probably closes #270 too.

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -1,16 +1,25 @@
 name: CompatHelper
 on:
   schedule:
-    - cron: '00 00 * * *'
+    - cron: 0 0 * * *
   workflow_dispatch:
 jobs:
   CompatHelper:
     runs-on: ubuntu-latest
     steps:
-      - name: Pkg.add("CompatHelper")
-        run: julia -e 'using Pkg; Pkg.add("CompatHelper")'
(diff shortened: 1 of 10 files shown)


sd109:
I've just noticed the new PR #306 which also makes some changes to the bloch_redfield_tensor function. None of the changes look like they conflict with each other, but will need to double check before merging this.

david-pl:
@sd109 Thanks for this! The changes look good to me. Is this ready to merge from your side? I will adapt #306 to handle conflicts.

sd109:
Yeah this is ready to merge as far as I'm concerned.

PR #312, opened 2021-07-15
Rename derivative functions

david-pl:


Krastanov:
Will these new `dsomething!` functions be made public API and/or exported?

david-pl:
@Krastanov They will be part of the API (and also of the documentation). As far as exporting goes, you'll have to pre-fix them with the submodule you're using (e.g. `timeevolution.dschroedinger!`), just as for the standard `timeevolution` functions.

PR #318, opened 2021-08-10
Return concise error message if Ket coefficients are not compatible with timeevolution methods

### LINKED ISSUE #317: Data type of `Ket()` sometimes incompatible with `timeevolution` methods
**The problem**
Whenever I define a basis state using the `Ket()` method I've noticed that the data type of the state vector is left to be the same as what I put in, see the following example:

    basis = SpinBasis(1//2)
    typeof( Ket(basis, [1, 0]).data )
        # = Array{Int64,1}
    typeof( Ket(basis, [1.0, 0.0]).data )
        # = Array{Float64,1}

I imagine it is maintained to use optimized methods for matrix multiplication and diagonalization, and most of the time it works great. When trying to solve time dependent problems I run into a problem, though. Trying the following, for instance

    basis = SpinBasis(1//2)
    H = sigmax(basis)
    ψ0 = Ket(basis, [1.0, 0.0])
    tspan = [0:0.01:1;]
    _, ψt = timeevolution.schroedinger(tspan, ψ0, H)

results in an `InexactError: Float64(0 - 1im)` with an enormous stacktrace that I have no hope of understanding. The problem goes away if I enforce the coefficients of the state vector to be complex, i.e. setting

    ψ0 = Ket(basis, complex([1.0, 0.0]))

I've run into this problem several times but keep forgetting how to solve it because the error message is opaque to me. Every time I figure it out I think: "Darn it, I should really remember this next time!" but that invariably doesn't happen.

**Proposed solution**
Would it be possible to have the `timeevolution` methods perform a check on the data type of the initial state, and have them convert it to complex if need be? Or at least throw an intelligible error message into the user's lap? In my view this would make the package more user friendly.

I would be happy to give this a go and file a pull request.

**Version data**
QuantumOptics v0.8.4
Julia 1.5.3

dehond:
Added a method `check_psi0` to `timeevolution_base` that uses Julia's type matching to check whether the coefficients of the initial wave function in time evolution methods are complex. If they are not, a short error message is generated that tells the user the coefficients should be converted to complex.

I've added this check to the `schroedinger`, `master`, and `mcwf` methods. During some unit tests, `schroedinger` methods are called with non-`Ket` datatypes as initial wave functions, so there's a generic `check_psi0` function that's called if the input is not of the `Ket` type. To make sure it passes the unit tests in `test_timeevolution_abstractdata.jl` I had to specify the input datatype to be
```Julia
check_psi0(psi0::Ket{B, T} where B<:Basis where T<:Array)
```
instead of just
```Julia
check_psi0(psi0::Ket)
````
In a sense the function is now checking the type twice, once to verify it's a `Ket{B, T}`, and then once more to make sure `T` is actually an array with complex coefficients. This seems suboptimal to me, but I couldn't immediately figure out a way to do it in one go. (For instance by having the value of the `T` that is matched in the function header available the function body.)

fixes #317

### DIFF
--- a/src/master.jl
+++ b/src/master.jl
@@ -192,7 +192,10 @@ end
 
 # Automatically convert Ket states to density operators
 for f ∈ [:master,:master_h,:master_nh,:master_dynamic,:master_nh_dynamic]
-    @eval $f(tspan,psi0::Ket,args...;kwargs...) = $f(tspan,dm(psi0),args...;kwargs...)
+    @eval function $f(tspan,psi0::Ket,args...;kwargs...)
+        check_psi0(psi0)
+        $f(tspan,dm(psi0),args...;kwargs...)
+    end
 end
 
 # Non-hermitian Hamiltonian
--- a/src/mcwf.jl
+++ b/src/mcwf.jl
@@ -457,6 +457,7 @@ Check input of mcwf.
 """
 function check_mcwf(psi0, H, J, Jdagger, rates)
     # TODO: replace type checks by dispatch; make types of J known
+    check_psi0(psi0)
     isreducible = true
     if !(isa(H, DenseOpType) || isa(H, SparseOpType))
         isreducible = false
--- a/src/schroedinger.jl
+++ b/src/schroedinger.jl
@@ -5,7 +5,8 @@ Integrate Schroedinger equation to evolve states or compute propagators.
 
 # Arguments
 * `tspan`: Vector specifying the points of time for which output should be displayed.
-* `psi0`: Initial state vector (can be a bra or a ket) or initial propagator.
+* `psi0`: Initial state vector (can be a bra or a ket) or initial propagator. Has to be
+        complex array.
 * `H`: Arbitrary operator specifying the Hamiltonian.
 * `fout=nothing`: If given, this function `fout(t, psi)` is called every time
         an output should be displayed. ATTENTION: The state `psi` is neither
@@ -15,6 +16,7 @@ Integrate Schroedinger equation to evolve states or compute propagators.
 function schroedinger(tspan, psi0::T, H::AbstractOperator{B,B};
                 fout::Union{Function,Nothing}=nothing,
                 kwargs...) where {B,T<:Union{AbstractOperator{B,B},StateVector{B}}}
+    check_psi0(psi0)
     dschroedinger_(t, psi, dpsi) = dschroedinger!(dpsi, H, psi)
     x0 = psi0.data
     state = copy(psi0)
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by bastikr** [comment]
I'm pretty sure this change is against what @david-pl tries to achieve - so please let's not merge this until he reviews this PR. Originally we took an approach more in the line of this PR, i.e. forcing many things to be of a specific type like states being stored as Vector{ComplexF64}. This has a few advantages like better error messages and guaranteeing that everything is reasonable fast since we can be sure the user does not use abstract types by accident (https://docs.julialang.org/en/v1/manual/performance-tips/#man-performance-abstract-container). However, the restriction to a specific type is something which is avoided in the Julia world since it harms flexibility.

**Comment by ChristophHotter** [comment]
@bastikr Okay, my first thought was also that it harms the flexibility but on the other hand it seemed to be very convenient to get a better error message in this special case. Anyway, I will discuss it with @david-pl.

**Comment by david-pl** [comment]
@dehond Sorry for the delay. As @bastikr said, I think implementing this change may harm flexibility, so I would argue against merging it. I will try to explain why:

Interoperability between Julia packages is based on using generic types. You already found one example of this, which is the `test_timeevolution_abstractdata.jl`, where `Ket`s are used that have data which acts like an array but isn't a subtype to `Array`. While there may not be an obvious use-case for this, it might come in handy when working with special data structures. An example at the level of `Operator` would be that one can use a `LinearMap` (from [LinearMaps.jl](https://github.com/Jutho/LinearMaps.jl)) as data which is useful sometimes.

While you took care of this properly by implementing the generic fallback method `check_psi(x)=nothing`, there could still be a case where one uses a special number type in the data array that is `<:Number`, but not `<:Complex`, while it still works with complex numbers using type promotion. In such a case, your check would falsely error. While I admit that this is quite a small edge case, which we could argue is not important, all we gain from this change is a different error message. That's why I would say it's probably not worth it.

Furthermore, the issue #317 all comes down to the following error
```julia
using LinearAlgebra
psi, dpsi = rand(5), rand(5)
H = rand(5,5)
mul!(dpsi,H,psi,-im,0)
```
which will reproduce your `InexactError`. The ugly stacktrace you report in #317 is because the error is thrown inside the time evolution. Long and incomprehensible stacktraces are an issue within Julia itself. If we start working around these ugly error messages, then where do we stop? Ugly stacktraces are thrown all the time due to comparably simple errors. I'd rather wait for a proper fix from the Julia developers to this issue.

I hope these explanations are sufficiently clear. In any case, thanks for the effort you put in this PR @dehond

dehond:
@david-pl thanks for your detailed explanation. I was worried this could be something that throws off a new user of what I feel is otherwise a very user-friendly and intuitive package. But if, as you point out, the problem is rather Julia's incomprehensible stack traces in general, it doesn't make much sense indeed to code around this for every corner case that does something like this. I can see how the resulting code will be hard to understand and maintain. Thanks for having a look at my PR, though!

PR #327, opened 2022-02-19
Use recursive formula for binomial coefficient in CSS construction

### LINKED ISSUE #326: Coherent spin states overflows for large spin number
~~~julia
julia> using QuantumOptics
julia> b = SpinBasis(35)
julia> css = coherentspinstate(b,0,0)
ERROR: OverflowError: binomial(70, 26) overflows
Stacktrace:
 [1] binomial(n::Int64, k::Int64)
   @ Base ./intfuncs.jl:1042
 [2] coherentspinstate(b::SpinBasis{35//1, Int64}, theta::Int64, phi::Int64)
   @ QuantumOptics ~/.julia/packages/QuantumOptics/pQzS7/src/phasespace.jl:299
 [3] top-level scope
   @ REPL[6]:1
~~~

The solution is to use a recursive definition of the binomial coefficient instead of the (integer) built-in `binomial` function. I'll prepare a PR for this.

goerz:
This uses a recursive formula for the binomial coefficient in the construction of a `coherentspinstate` in order to avoid the overflow associated with the built-in `binomial` for larger values of S

Closes #326

### DIFF
--- a/src/phasespace.jl
+++ b/src/phasespace.jl
@@ -288,19 +288,31 @@ parametrization is not unique), similarly to a qubit on the
 Bloch sphere.
 """
 function coherentspinstate(b::SpinBasis, theta::Real, phi::Real)
+
     result = Ket(b)
     data = result.data
+
     N = length(b)-1
-    sinth = sin(0.5theta)
-    costh = cos(0.5theta)
-    expphi = exp(0.5im*phi)
-    expphi_con = conj(expphi)
-    sb = 1.0  # = recursive definition of √(N over n)
-    @inbounds for n=0:N
-        data[n+1] = sb * (sinth*expphi)^n * (costh*expphi_con)^(N-n)
-        sb = sb * √((N - n) / (n + 1))  # N over n = (N+1-n)/n (N over n-1)
+    α = sin(theta / 2) * exp(1im * phi / 2)
+    β = cos(theta / 2) * exp(-1im * phi / 2)
+
+    # forward pass: `c_n = sqrt(binomial(N, n)) * α^n` with `n  ≥ 0`,
+    # using recursive `binomial(N, n) = ((N+1-n)/n) * binomial(N, n-1)`
+    coefficient = 1.0
+    @inbounds for n = 1:N+1
+        data[n] = coefficient
+        coefficient *= α * sqrt((N + 1 - n) / n)
     end
+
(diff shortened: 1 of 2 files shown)


goerz:
Actually, you could still make this faster by a factor of 2-8 (depending on the size of the Hilbert space) and numerically more stable by using recursion not just of the binomial coefficient, but also for the rest of the equation. I can add another commit for that...

goerz:
Ok, should be ready now.

The performance benefit might not seem very relevant just of initializing a state, but it really speeds things up when you use the routine for creating the usual visualizations (especially animations) of spin-squeezed states like this:

spinsqueezed

where every point on the sphere is the overlap with the CSS pointing in that direction.

david-pl:
@goerz Thanks for the PR! There's actually a duplicate issue here: https://github.com/qojulia/QuantumOpticsBase.jl/issues/43, which is also solved by your changes.

saulpila:
By the way, now that we are talking about  `coherentspinstate` efficiency, if you additionally change

https://github.com/qojulia/QuantumOptics.jl/blob/871183ffbec52f0a41af255415283a6c7206a24f/src/phasespace.jl#L290-L293

to

    function coherentspinstate(b::SpinBasis, theta::Real, phi::Real, result::Ket=Ket(b))
            data = result.data

Then it allows to reuse the same array when doing visualizations of the Husimi function and prevents excessive memory allocation.  This should speed things up in e.g. `qfuncsu2` by changing

https://github.com/qojulia/QuantumOptics.jl/blob/871183ffbec52f0a41af255415283a6c7206a24f/src/phasespace.jl#L338-L340

to

     coherent = Ket(b) 
     @inbounds  for i = 0:Ntheta-1, j = 0:Nphi-1 
         coherentspinstate(b,pi-i*pi/(Ntheta-1),j*2pi/(Nphi-1)-pi,coherent)
         result[i+1,j+1] = (2*lb+1)/(4pi)*abs2(psi_bra_data*coherent.data) 
     end

goerz:
I agree that could be useful, but it should probably just be a `coherentspinstate!` function, like so:

```julia
coherentspinstate!(Psi::Ket, theta::Real, phi::Real)
```

where the basis is extracted from `Psi`.

Then, `coherentspinstate` can delegate to `coherentspinstate!`.

PR #337, opened 2022-06-13
Force real probabilities to avoid inexact error

### LINKED ISSUE #335: Inexact error Complex -> Float with MCWF jumps
Hi, came across an inexact error simulating with `mcwf_dynamic`;
 
I think in this line I get very small imaginary parts from `dot(psi_new.data, psi_new.data)`
https://github.com/qojulia/QuantumOptics.jl/blob/b2618ee33c9639617bcba11add5484052efa552b/src/mcwf.jl#L397

Defining a more forgiving conversion seems to solve all issues:
`Base.convert(::Type{T}, x::Complex) where {T<:AbstractFloat} = abs(imag(x)) < 1e-16 ? T(real(x)) : throw(InexactError(nameof(T), T, x))`

But taking the `real` part, or the `norm` instead of `dot` is probably more efficient.

I've put together a small example Pluto notebook, but seems l can only upload the pdf:
[🎈 kpo_mcwf_mwe.jl — Pluto.jl.pdf](https://github.com/qojulia/QuantumOptics.jl/files/8676372/kpo_mcwf_mwe.jl.Pluto.jl.pdf)

david-pl:
Fixes #335


david-pl:
We need the fix here https://github.com/SciML/DiffEqBase.jl/pull/769 for builds to pass.

PR #340, opened 2022-08-25
added method for recast!

AmitRotem:
dynamics of Operator with different left and right basis would fail.
for example;
```
b1 = FockBasis(4)
b2 = NLevelBasis(2)
H(t,u) = number(b1) + destroy(b1) + create(b1)
ψ0 = projector(b1, SubspaceBasis(b1,[basisstate.([b1],1:2)...]))
ψ1 = Operator(ψ0.basis_l, FockBasis(1), ψ0.data)
timeevolution.schroedinger_dynamic(range(0,1,10), ψ1, H); # ok
timeevolution.schroedinger_dynamic(range(0,1,10), ψ0, H); # fail
```

`recast!` of an Operator type with different left and right basis fixes this.
This is useful for propagating a subset of states.

### DIFF
--- a/src/schroedinger.jl
+++ b/src/schroedinger.jl
@@ -56,6 +56,10 @@ a vector or a matrix with a matching size.
 """
 recast!(psi::StateVector{B,D},x::D) where {B, D} = (psi.data = x);
 recast!(x::D,psi::StateVector{B,D}) where {B, D} = nothing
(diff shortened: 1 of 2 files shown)


amilsted:
Ah, I think I've seen this before. Thanks Amit!

amilsted:
We should add a simple test of this case.

AmitRotem:
I've added a test

amilsted:
LGTM

PR #344, opened 2022-10-21
Allow supplying jump state.

amilsted:


src/mcwf.jl
```diff
@@ -342,15 +349,39 @@ function integrate_mcwf(dmcwf, jumpfun, tspan,
...
+Stores the RNG used to generate jump thresholds, as well as
+the most recent threshold rolled. A jump is carried out when
+the norm-squared of the evolving state drops below the threshold.
+
+Can be passed to `mcwf()` and related functions as the `rng_state`
+keyword argument to persist the state across calls.
+"""
+mutable struct JumpRNGState{R<:AbstractRNG}
+    rng::R
+    threshold::Float64
```
david-pl, line 364: Should `threshold` have a type parameter? Might be useful to e.g. have this as `Float32` when running things on GPU.

PR #356, opened 2022-12-28
Promote state and time-span in `schroedinger` to the type from `H(t)*psi`

AmitRotem:
For optimizing `cost` wrt `par`, it is sometimes useful to get the gradient of `cost` wrt `par`.
Promoting the state and time-span to the type from `Hamiltonian * state` makes it possible to use `ForwardDiff.jl` on `cost` - if you happen to define the Hamiltonian in a compatible manner to `ForwardDiff.jl`.

For example;
```
using QuantumOptics
import Preferences, UUIDs
Preferences.set_preferences!(UUIDs.UUID("f6369f11-7733-5829-9624-2563aa707210"), "nansafe_mode" => true)
import ForwardDiff as FD

# 3 level kerr transmon with drive
ba = FockBasis(2)
T2 = 1e4
function get_Ht(p::Vector{<:Tp}) where Tp
    ω0, α = Tp(5.0), Tp(-0.2)
    A, freq, ϕ, T = p
    op = 2π*([number(ba), 2\create(ba)*number(ba)*destroy(ba), im*(create(ba)-destroy(ba))])
    fω(t) = ω0
    fα(t) = α
    fΩ(t) = A*cospi(2t*freq + 2ϕ)*sinpi(t/T)^2
    H_at_t = LazySum(zeros(Tp,length(op)), op)
    function Ht(t,_)
        H_at_t.factors.= (fω(t), fα(t), fΩ(t))
        return H_at_t
    end
    return Ht
end

ψ0 = Operator(SpinBasis(1/2), basisstate(ba, 1), basisstate(ba, 2))
target = ψ0*exp(im*0.5π*dense(sigmax(SpinBasis(1/2)))) # x gate
function cost(par)
    T = last(par)
    Ht = get_Ht(par)
    ts = (0.0, T)
    _, ψT = timeevolution.schroedinger_dynamic(ts, ψ0, Ht)
    1-abs2(tr(target'last(ψT))/2)*exp(-T/T2)
end

p0 = [0.03, 5.0, 0.25, 100.0]
cost(p0) |> display
FD.gradient(cost, p0)
```


Krastanov:
Could you remove Manifest.toml from this pull request?

AmitRotem:
For this small example;
"""
p0 = [0.03, 5.0, 0.25, 100.0]
cost(p0);
@benchmark cost(p0)
"""
In this branch, benchmarking gives run time of 21.429 ms ± 797.528 μs with 289 allocations.
In master branch, benchmarking gives run time of 22.229 ms ±  1.182 ms with 277 allocations.
So mostly a small allocation overhead ?

I'm guessing this will be true for the general case of working with ComplexF64.
Do you think there is some edge cases where the promotion will results in some abstract type, resulting lots of allocations ?
This happens in this example when taking the gradient if I don't promote ω0 or α.

src/schroedinger.jl
```diff
@@ -102,3 +104,18 @@ function check_schroedinger(psi::Bra, H)
...
     check_samebases(H)
 end
+
+
+_promote_time_and_state(tspan, psi0, f) = _promote_time_and_state(tspan, psi0, f(first(tspan), psi0))
+function _promote_time_and_state(tspan, psi0, H::AbstractOperator)
+    # general case is Ts<:Complex, Tt<:Real
+    Ts = eltype(H)
+    Tt = real(Ts)
+    (isconcretetype(Ts) && isconcretetype(Tt)) || @warn "For using `ForwardDiff` on `schroedinger` the element type of `real(H(t,psi)*psi)` must be concrete !!\nGot elements of type $Tt \nTry promoting the Hamiltonian elements based on the parameters you are differentiating by."
```
Krastanov, line 114: In what situations would this warning message be seen? I feel it might be confusing as it will show up in front of users that have not touched `ForwardDiff` or `schroedinger`.

src/schroedinger.jl
```diff
@@ -102,3 +104,18 @@ function check_schroedinger(psi::Bra, H)
...
+    tspan = Tt.(tspan)
```
Krastanov, line 115: I imagine there are situations where no promotion is necessary. Can we shortcircuit those?

Krastanov:
The extra allocation does not seem particularly problematic on first sight (as long as it is just a constant non-growing overhead), but it can probably be removed with a bit of extra logic in the promotion function.

But I am a bit confused, why is promotion even necessary. You are not deriving with respect to `state` or `tspan`, so why is it necessary to make them dual numbers? I thought ForwardDiff can multiply by a constant? This works for instance without explicitly promoting the vector `[1,2,3]`:

```
julia> cost(H) = sum(H*[1.,2.,3.])

julia> ForwardDiff.gradient(cost, rand(3,3))
...
```

Krastanov:
I guess the answer to my question about why promotion is necessary is because there are in-place operations done by the diff eq solver?

How about we just shortcircuit on the `isconcrete` checks. If the types are concrete and if they are different from the types of tspan and state, then we modify them. If not, we do not touch them?

Krastanov:
I am also a bit worried: how does this work when another autodiff tool is used? Is this making Zygote and Enzyme happier or is it breaking things for them.

src/schroedinger.jl
AmitRotem: I'll remove it.
This was trying to warn about something like
```
ω0, α = 5.0, -0.2 # not promoting
#...
fun = [fω, fα, fΩ]
H_at_t = LazySum([f(0.0) for f=fun], op)
@show H_at_t.factors #  type is abstract !! Complex{Real}
```
in the definition of `get_Ht`.
But It doesn't cover all possible "bad" definitions of `get_Ht`.

src/schroedinger.jl
AmitRotem: I'll declare type `Dual` and only promote that case, probably would be safer.

AmitRotem:
> s the answer to my question about why promotion is necessary is because there are in-place operations done by the diff eq solver?

I think it's a QO.jl issue, meaning in-place operations on states and operators. Using DiffEq.jl directly with matrix and vector there no need for the user to promote anything.

Krastanov:
> > s the answer to my question about why promotion is necessary is because there are in-place operations done by the diff eq solver?
> 
> I think it's a QO.jl issue, meaning in-place operations on states and operators. Using DiffEq.jl directly with matrix and vector there no need for the user to promote anything.

I think the correct solution would be to fix that issue in QO.jl instead of working around it. My worry is that workarounds that do not follow the DiffEq.jl style would just become more and more difficult to keep functional as the autodiff capabilities in julia evolve.

Any idea why DiffEq does not have this problem, but QO has it? Anything in particular that QO does that makes things difficult?

AmitRotem:
> I am also a bit worried: how does this work when another autodiff tool is used? Is this making Zygote and Enzyme happier or is it breaking things for them.

Last time I tryed Zygote didn't work with QO, and I didn't check if the promoting changed anything.
Anyway, If we promote only if H is Dual it shouldn't affect other packages right?

Krastanov:
> > I am also a bit worried: how does this work when another autodiff tool is used? Is this making Zygote and Enzyme happier or is it breaking things for them.
> 
> Last time I tryed Zygote didn't work with QO, and I didn't check if the promoting changed anything. Anyway, If we promote only if H is Dual it shouldn't affect other packages right?

That is true, but similarly to my comment above, this would be a fragile fix around some other underlying issue. I would be hesitant to use a library that has such special cases in its code, and such a library would be more difficult for others to continue developing. Moreover, that approach would need to add ForwardDiff as a dependency.

Figuring out why QO does not support something that is supported by the underlying DiffEq would probably lead to simpler code and support for other unrelated autodiff tools. Could you give a minimum example of a working autodiff with DiffEq?

AmitRotem:
> > > s the answer to my question about why promotion is necessary is because there are in-place operations done by the diff eq solver?
> > 
> > 
> > I think it's a QO.jl issue, meaning in-place operations on states and operators. Using DiffEq.jl directly with matrix and vector there no need for the user to promote anything.
> 
> I think the correct solution would be to fix that issue in QO.jl instead of working around it. My worry is that workarounds that do not follow the DiffEq.jl style would just become more and more difficult to keep functional as the autodiff capabilities in julia evolve.
> 
> Any idea why DiffEq does not have this problem, but QO has it? Anything in particular that QO does that makes things difficult?

I think it's the preallocation of `dpsi`, than `mul!` is failing with ForwardDiff because it try to convert Dual from `H*psi` to ComplexF64 in `dpsi` - related to https://github.com/qojulia/QuantumOpticsBase.jl/issues/76 .

DiffEq.jl Also had this issue in the past, meaning the user had to promote time and state for ForwardDiff to work. Not sure what solution they implemented internally.

Krastanov:
I think the fix should be in `integrate` or `recast!` where we are interacting with the low level DiffEq library.

https://github.com/qojulia/QuantumOptics.jl/blob/master/src/timeevolution_base.jl#L14

https://github.com/qojulia/QuantumOptics.jl/blob/master/src/schroedinger.jl#L57

Maybe we should try to expose some of SciMLSensitivity https://docs.sciml.ai/SciMLSensitivity/stable/

Krastanov:
This might be of use: https://github.com/SciML/PreallocationTools.jl

AmitRotem:
> > > I am also a bit worried: how does this work when another autodiff tool is used? Is this making Zygote and Enzyme happier or is it breaking things for them.
> > 
> > 
> > Last time I tryed Zygote didn't work with QO, and I didn't check if the promoting changed anything. Anyway, If we promote only if H is Dual it shouldn't affect other packages right?
> 
> That is true, but similarly to my comment above, this would be a fragile fix around some other underlying issue. I would be hesitant to use a library that has such special cases in its code, and such a library would be more difficult for others to continue developing. Moreover, that approach would need to add ForwardDiff as a dependency.
> 
> Figuring out why QO does not support something that is supported by the underlying DiffEq would probably lead to simpler code and support for other unrelated autodiff tools. Could you give a minimum example of a working autodiff with DiffEq?
Cheating a bit, because there is an issue with complex numbers... "Dual on Complex is ambiguous".
```
using LinearAlgebra, DifferentialEquations, SciMLSensitivity
import ForwardDiff as FD
import Zygote as Zy

H0 = randn(5,5)
H0-=H0'
A = randn(5,5)
A-=A'
u0 = hcat(normalize(randn(5)), normalize(randn(5)))

function Ht!(du,u,p,t)
    a,b,c = p
    du .= A*u
    du.*= a*cos(b*t+c)
    du.+= H0*u
    nothing
end

function loss(p)
    prob = ODEProblem(Ht!,u0,(0.0,1.0),p)
    sol = solve(prob)
    abs2(tr(u0'last(sol.u)))
end

loss(rand(3))
FD.gradient(loss, rand(3))
Zy.gradient(loss, rand(3))
```

AmitRotem:
Not sure if there is a nice way to deal with complex number here.
https://docs.sciml.ai/SciMLSensitivity/dev/optimal_control/SDE_control/ here they split the eq to real and imag parts.

If everything is nice and analytic, I think it's safe to define Dual on Complex;
```
function FD.Dual{T, V, N}(Z::Complex) where {T, V, N}
    complex(FD.Dual{T, V, N}(real(Z)), FD.Dual{T, V, N}(imag(Z)))
end
```
then we only need to promote the time-span, and the above DiffEq.jl example works with complex numbers.

I'm not sure how or why this issue do not show up with QO.jl

AmitRotem:
I will look into using PreallocationTools.jl to better handle edge cases

Krastanov:
I tried to read through the DiffEq code, and it seems they do the promotions themselves already:

https://github.com/SciML/DiffEqBase.jl/blob/82a69aecf6f20062658f6a554695b56821f4f25d/src/solve.jl#L936

And they do the promotion in a way that is very similar to what you have created here:

https://github.com/SciML/DiffEqBase.jl/blob/aad604cc44d9ae2584a94936b8c8eddc29293662/src/forwarddiff.jl#L207

It would be fantastic if we can figure out why the promotion does not work by default for us. That way we would not need to add extra code to QO.jl and we will have it for free for all dynamical equations, not just the `shroedinger` one.

Krastanov:
The DiffEq promotion does not automatically work, because their promotion happens only if there is a dual in the `p` parameters argument. We never use the parameters argument, so it never gets triggered. Here is an example of this not working that does not involve QuantumOptics:

```
function make_f(param)
    function fun!(du,u,p,t)
        du .+= param .* u
    end
end
function cost(param)
    fun! = make_f(param)
    prob = ODEProblem(fun!,[1.0,0.0],(0,.1))
    state = solve(prob,Tsit5(),save_on=false,save_end=true,save_start=false,wrap=false)[end]
    sum(state)
end

cost(1.0)
ForwardDiff.derivative(cost,1.0)  # breaks
```

Krastanov:
My personal belief is that the "proper" solution is to rework a lot of the internals of QO.jl in order to use DiffEq.jl in a more idiomatic manner. That would significantly simplify the code.

However, that is a gigantic amount of work that I do not think should be blocking important features like the one you are currently suggesting.

Therefore, I am now in favor of the original approach you had started with, even if it re-implements features that supposedly we should get from DiffEq for free. I would suggest using a check similar to `eltype(...) <: ForwardDiff.Dual` so that promotion is veeeeeeeeery limited.

AmitRotem:
> The DiffEq promotion does not automatically work, because their promotion happens only if there is a dual in the `p` parameters argument.
> ....

This reminds me of how to get ForwardDiff.jl working on an ode with complex numbers;
```
function loss(p)
    Tp = eltype(p)
    Ts = promote_type(Tp, eltype(u0))
    function f(u,qwe,t)
        Ht(u,p,t)
    end
    prob = ODEProblem( f, Ts.(u0), Tp.((0.0,1.0)) )
    sol = solve(prob)
    abs2(tr(first(sol.u)'last(sol.u)))
end
```
This bypass the promotion in DiffEqBase.jl using the type from `p`, as you mentioned.
I'm not sure `DiffEqBase.promote_u0` fails on complex.

I've opened an issue https://github.com/SciML/DiffEqBase.jl/issues/858

Krastanov:
- [ ] is it necessary to promote `tspan`?
- [ ] could we add some simple tests? Maybe a whole new file given that autodiff has never been tested before. The quote below is a rough suggestion for something that tests both the static and dynamic solvers
```
using QuantumOptics
using ForwardDiff

ba = FockBasis(5)
ψ0 = basisstate(ba, 1)
target = basisstate(ba, 2)
function get_Ht(p)
    op = [create(ba)+destroy(ba)]
    f(t) = sin(p*t)
    H_at_t = LazySum([f(0)], op)
    function Ht(t,_)
        H_at_t.factors .= (f(t),)
        return H_at_t
    end
    return Ht
end

function get_H(p)
    op = create(ba)+destroy(ba)
    return sin(p)*op
end

function cost1(par)
    Ht = get_Ht(par)
    ts = (0.0, 1.0)
    _, ψT = timeevolution.schroedinger_dynamic(ts, ψ0, Ht)
    abs2(target*last(ψT))
end

function cost2(par)
    H = get_H(par)
    ts = (0.0, 1.0)
    _, ψT = timeevolution.schroedinger(ts, ψ0, H)
    abs2(target.data'*last(ψT))
end

cost1(1.0)
cost2(1.0)
ε = 1e-9
ForwardDiff.derivative(cost1,1.0) ≈ (cost1(1.0+ε)-cost1(1.0))/ε
ForwardDiff.derivative(cost2,1.0) ≈ (cost2(1.0+ε)-cost2(1.0))/ε
```

src/schroedinger.jl
```diff
@@ -102,3 +104,21 @@ function check_schroedinger(psi::Bra, H)
...
     check_samebases(H)
 end
+
+
+_promote_time_and_state(tspan, psi0, f) = _promote_time_and_state(tspan, psi0, f(first(tspan), psi0))
+function _promote_time_and_state(tspan, psi0, H::AbstractOperator)
+    Ts = eltype(H)
+    Tt = real(Ts)
+    # promote only if ForwardDiff.Dual
+    if Tt <: ForwardDiff.Dual
```
Krastanov, line 114: could you add an extra condition, that psi0 is not already dual? We do not want to apply `Dual` twice.

AmitRotem: done.
Also added check to promote if either state or time are `Dual `

AmitRotem:
> * [ ]  is it necessary to promote `tspan`?

yes, not sure why. DiffEqBase.jl also promote tspan.

AmitRotem:
If the `Dual` on  `Complex` thing will be resolved in DiffEqBase.jl we could pass an empty `p` vector of `Dual` to get the promotion done in DiffEqBase

test/test_ForwardDiff.jl
```diff
@@ -0,0 +1,125 @@
...
+function cost01(par)
+    Ht = getHt(par)
+    ts = (0.0, 1.0)
+    _, ψT = timeevolution.schroedinger_dynamic(ts, psi, Ht)
+    abs2(target0'*last(ψT))
+end
+
+cost01(rand())
+FD.derivative(cost01, rand())
+@test all([isapprox(FD.derivative(cost01, 1.0), (cost01(1.0+ε)-cost01(1.0))/ε, atol=1e-7) for q=range(0,2π,2^7)])
```
Krastanov, line 43: what is the purpose of `for q=...`?

Krastanov:
I like the style randomized tests, but I am noticing that this adds 1m30s to the test duration. Maybe we can test over 2 or 3 random samples instead of over 2^7?

Krastanov:
Some of the repeated tests do not seem to actually be random though. For those I would suggest just doing one single test.

test/test_ForwardDiff.jl
AmitRotem: should have evaluate at `q`

AmitRotem:
> I like the style randomized tests, but I am noticing that this adds 1m30s to the test duration. Maybe we can test over 2 or 3 random samples instead of over 2^7?

it was mostly the 3'rd test with large tspan. reduced it, and the repetitions

Krastanov:
This seems pretty reasonable to me and it is functionality I would love to have. I would be in favor of merging after some minor cleanup:

- Removing the local preferences toml file
- Removing the `@time` annotations (to make the test report cleaner)

Maybe adding an issue on the issue tracker saying something along the ways of "adding ForwardDiff promotions to all solvers" (we have quite a bit more than just Schroedinger) with a link to this pull request. It could be a good test issue for google summer of code applicants.

Krastanov:
Instead of working with a preferences toml file, is it not possible to just set it from code: 

```
julia> using ForwardDiff, Preferences

julia> set_preferences!(ForwardDiff, "nansafe_mode" => true)
```

As seen here: https://juliadiff.org/ForwardDiff.jl/stable/user/advanced/#Fixing-NaN/Inf-Issues

AmitRotem:
Yes, but then don't we need preferences package in the project file?

Krastanov:
It is needed only as a test dependency.

However, do tests actually fail with the standard NaN mode? That would be a bit worrisome on its own. If that happens often, it should probably also be explained in the documentation.

AmitRotem:
I believe the NaN thing also appears in DiffEqBase.jl

What do you think about switching from testing vs finite diff to testing vs ForwardDiff on DiffEq, (without QuantumOptics.jl) ?

Krastanov:
> I believe the NaN thing also appears in DiffEqBase.jl (for ForwardDiff with complex numbers ode)

I am not sure I understand. I thought NaNs should happen when there is something unstable or "singular" about the numerics. Are dual complex numbers inherently causing it?

> 
> What do you think about switching from testing vs finite diff to testing vs ForwardDiff on DiffEq, (without QuantumOptics.jl) ? This will get around all these tolerance issues.

I do not think that would fix tolerance issues, just hide them. QuantumOptics uses different default solver, different time stepping, etc, so I would be surprised if the answers are numerically equal. Isn't it easier just to set a very loose tolerance, e.g. 1e-4?

AmitRotem:
> I am not sure I understand. I thought NaNs should happen when there is something unstable or "singular" about the numerics. Are dual complex numbers inherently causing it?

I don't think it's dual complex numbers issue, but maybe something in the algorithm that gives the initial time step (I'm basing this hunch on a simple example for getting NaN and how to "resolve" them https://github.com/SciML/DiffEqBase.jl/pull/860#issuecomment-1370350444)
Here in the test I used some initial time step to avoid the NaN's.
 
> I do not think that would fix tolerance issues, just hide them.

Yes, but, hide them as in acknowledging that tolerance of Dual number are something DiffEq should deal with.

> QuantumOptics uses different default solver, different time stepping, etc, so I would be surprised if the answers are numerically equal. Isn't it easier just to set a very loose tolerance, e.g. 1e-4?

The test that failed was due to one instance of 1.3e-3 abs diff, other tries passed with abstol of 1e-6.
We can compare with the same solver, say Vern8, and saving only the final state. Not sure what other options will matter here.

test/test_ForwardDiff.jl
```diff
@@ -0,0 +1,221 @@
...
+p0 = [0.3, ω00, 0.25, 10.0]
+rp(k) = p0 .* ( ones(length(p0))+k*(1 .-2rand(length(p0))) )
+### check that nothing fails
+rp(rand())
+cost1(p0)
+FDgrad(cost1, p0)
+FDgrad(cost1_with_dt, p0)
+### test vs finite difference
+#### there somtimes an issue with the gradient at p0, which is alot less accurate
+@test all([test_vs_fin_diff(cost1_with_dt, p; atol=1e-6) for p=rp.(range(1e-9, 0.1, tests_repetition))])
```
AmitRotem, line 176: There is a small unstable region around `p0`, where AD and finite diff differ by `1e-2`. Not sure why that is.

Krastanov: Would just removing that test make sense? Is it really testing anything that the previous set of much simpler tests do not? This is about testing the autodiff, not about a study of numerical stability.

AmitRotem: Yes.
I didn't check this example vs DiffEq, but given that the other examples I compared showed less than 1e-12 difference I think we can remove this test.
And also the 1'st test?

Krastanov: I am not sure which one you refer to as 1st. Testing both the static and dynamic Hamiltonians seems valuable to me.

AmitRotem: Sure

test/test_ForwardDiff.jl
```diff
@@ -0,0 +1,221 @@
...
+cost01(rand())
+FDgrad(cost01, rand())
+fin_diff(cost01, rand())
+### test vs finite difference
+@test all([test_vs_fin_diff(cost01, q; atol=1e-7) for q=vcat(0,π,rand(tests_repetition)*2π)])
+
+## static
+function get_H(p)
+    op = create(ba0)+destroy(ba0)
+    return sin(p)*op
```
AmitRotem, line 83: I mean keep this static test

test/test_ForwardDiff.jl
```diff
@@ -0,0 +1,221 @@
...
+# ex2
+ba2 = FockBasis(3)
+A, B = randoperator(ba2), randoperator(ba2)
+A+=A'
+B+=B'
+ψ02 = randstate(ba2)
+target2 = randstate(ba2)
+function cost2(par)
+    a,b = par
+    Ht(t,u) = A + a*cos(b*t)*B/10
```
AmitRotem, line 189: And this dynamic test

test/test_ForwardDiff.jl
Krastanov: Keeping the two tests you specifically cited below seems reasonable to me. You can also guide yourself by the coverage report - we should aim to have all new lines covered by tests.

AmitRotem:
Are we happy with this?

Krastanov:
I think the way it is done is quite clean, or at least as clean as possible.

I am not the biggest fan of the style of the tests: my personal philosophy is to not include tests that are sensitive to numerical instabilities (if the code being tested is unrelated to numerical instability control). But being practical, I do not think it is necessary to redo the tests or block the merge. Maybe in a couple of years there will be a test failure due to some change to default DiffEq / ForwardDiff algorithms, but that is fine.

I am a bit worried though that this adds whole 3 minutes to the test run. I would have expected testing this to take 10-20 seconds. Wouldn't both these issues (instabilities and long test runs) be fixed if we just test fewer examples. Testing fewer examples would not drop the test coverage percentage or really change much.

Minor notes:
- There is not much point in seeding the random number generator: the generator algorithm is not stable between versions and architectures. If you insist on it, you can use StableRNG.jl but that seems like overkill
- The comments in the test file were a bit disorganized, e.g. it jumps from Example 0 to Example 2.

Anyway, I do not see blocking issues. @david-pl , is there a policy/philosophy about what the tests need to look like. These seem good enough even if unpolished and a bit long.

More importantly, thanks for fixing some of the ergonimics with autodiff in this package. This would be much more pleasant to use for me now.

david-pl:
@AmitRotem thanks for all the work here! I have to admit that my knowledge of AD is a bit limited, so I trust @Krastanov here - thanks for the detailed review!

Regarding the tests: they are a bit of a mess throughout the package already. Also, time is a bit of an issue as the CI already takes quite long, but we'll have to look at that in a separate PR. There's no real philosophy in how to do things, except maybe that we have one test per file. So long as there are tests for new code, I'm happy!

Any mess we have in the tests can be cleaned up in the future. For that I'd simply stick to the way SciML tests are structured.

So all good from my side. @Krastanov just give me one last go ahead once you're happy with the PR and I'll merge it.

Krastanov:
@AmitRotem , if I understand correctly, this depends on a change you contributed to DiffEqBase in version [6.113.0](https://github.com/SciML/DiffEqBase.jl/releases/tag/v6.113.0). Could you add this to the Project dependencies and compat? You can also then directly do `using DiffEqBase` instead of getting it through the `OrdinaryDiffEq` namespace.

@david-pl  good to merge in my opinion after adding the dependency. Details below.

The comments about the tests stand, but as you mentioned that requires an overall cleanup, probably following SafeTestsets as done in SciML.

This also introduces a dependence on a private DiffEqBase function (the promotion one), however that function has existed since 2018 and very early versions of DiffEqBase. We can not really rely on semver to detect issues with this, but it seems relatively safe (and testing will detect any possible issues in the future). Amit has added another method to that function upstream to make sure it does not break on complex numbers.

As I already mentioned. The actual changes look neat and clean. Once this is merged I will create issues to track the addition of the same features to the rest of the solvers.

AmitRotem:
Most of the time is compilation time for ForwardDiff.jl.
Even if I keep only 2 test (not repeating over randoms), for checking no error during promotion and comparing AD on QO.jl vs AD on DiffEq.jl (which i think is the bare minimum to test for), the run time (localy) goes from 77 sec to 34 sec.

Krastanov:
Looks good to merge.

I do not actually think there is that much need for the tests against DiffEq. We are using DiffEq internally anyway, so just checking for correctness with a quick finite differences approximation on a numerically stable example would also have been enough.

AmitRotem:
If you think this minimal version of the test is enough, I'll remove the longer version.

Krastanov:
Thanks for the test cleanup! Looks good to merge to me (the only changes from the previous time I said that is that tests were streamlined).

PR #365, opened 2023-06-19
add a simple extra test to the steady state solver

Krastanov:


Krastanov:
I needed this test as I was debugging https://github.com/qojulia/QuantumOpticsBase.jl/issues/106 -- it is nothing sophisticated, just making the already present tests a bit more explicit

PR #371, opened 2023-07-28
Add const checking for non-dynamic evolution

amilsted:


Krastanov:
May we make the error message a bit more detailed? I am imagining an unexperienced undergrad that has not read the documentation very thoroughly. Maybe something along the lines of:

"You are attempting to use a time-dependent dynamics generator (a Hamiltonian or Lindbladian) with a solver that assumes constant dynamics. To avoid errors, please use the `_dynamic` solvers instead, e.g. `schroedinger_dynamic` instead of `schroedinger`."

Longer term, I am wondering why is there an error at all if we already know what should actually happen. Should we just dispatch to the correct solver if we detect `! is_const` (with maybe a `@warn` warning)? No need to fix this here, but it can maybe be posted on the issue tracker and marked as "good first issue"?

Otherwise, looks good to me (presumably after the is_const fixes get merged to QOBase.jl).

amilsted:
Yeah, I agree we should probably just "do the right thing" longer term and not require the "dynamic" methods. We could just add appropriate `set_time!()` to the solvers everywhere, which would compile out for common const cases. Or just branch on `is_const`, as you suggest.

amilsted:
Should be ready now @Krastanov @ChristophHotter

Krastanov:
LGTM. I assume the repeated definition of `_check_const` is just to avoid mixing namespaces/modules? I am fine with it if it is on purpose.

amilsted:
Yeah, that's the only reason.

PR #372, opened 2023-08-16
squeeze operators

karolpezet:
squeeze operator for harmonic oscillator and collective spins + tests

### DIFF
--- a/src/phasespace.jl
+++ b/src/phasespace.jl
@@ -558,3 +558,34 @@ function _calc_ylm_norm(l, m)
     end
     return n
 end
+
+"""
+    squeeze(b::FockBasis, x::Complex)
+
+This function generates a squeeze operator with squeezing parameter |x| 
+along squeezing direction specified by arg(x)/2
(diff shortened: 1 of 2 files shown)


Krastanov:
Any specific reason this submitted to QuantumOptics.jl? I think it is already implemented in QuantumOpticsBase.jl https://github.com/qojulia/QuantumOpticsBase.jl/blob/master/src/fock.jl#L70

I had the impression that QuantumOptics.jl is for the dynamics, while QuantumOpticsBase.jl is for this lower level operator constructs.

ChristophHotter:
@Krastanov Yes you are completely right. 
@karolpezet I think the squeeze operator for the SpinBasis is not implemented. Would you like to make a PR to QuantumOpticsBase.jl for this?

karolpezet:
oh yeah I missed the definition of squeeze operator. I did not find it in docs and assumed it is absent. I will fix it by moving it to QuantumOpticsBase.jl

Issue #376, opened 2023-12-16
ForwardDiff fails on (schroedinger_dynamic with) TimeDependentOperator


AmitRotem:
derivative on `TimeDependentSum` fails;
```
using QuantumOptics
import ForwardDiff as FD
base=SpinBasis(1/2)
ψi = spinup(base)
ψt = spindown(base)

function Ftdop(q)
    H= TimeDependentSum([q, abs2∘sinpi], [sigmaz(base), sigmax(base)])
    _, ψf = timeevolution.schroedinger_dynamic(range(0,1,2), ψi, H)
    abs2(ψt'last(ψf))
end
Ftdop(1.0)
FD.derivative(Ftdop, 1.0) # fail
```
`schroedinger_dynamic` promotes `tspan` and `state`  to `Dual` when needed, then `TimeDependentSum` is called with a `Dual` and fails.

Possible to call `_promote_time_and_state` here;
https://github.com/qojulia/QuantumOptics.jl/blob/e593ae1a51ebc47661322e8ac2c0116085708b2a/src/schroedinger.jl#L57-L60
and then promote `H` as `TimeDependentSum(op.coefficients, op.static_op; init_time=zero(promote_type(eltype(tspan), typeof(current_time(op)))))`
Or somehow pass this to `schroedinger_dynamic_function` ?

amilsted:
Hi Amit! Right, so the problem is indeed the type of the `current_time` field. We might want it to be a union including the appropriate Dual type here? Does that work if you do it in your `Ftdop` above?

Actually, I'm a bit puzzled in this case, as you are not taking a derivative wrt. time. Perhaps you could post the stack trace?

AmitRotem:
Hi Ash! I hope you're doing well and having fun!
`schroedinger_dynamic` is promoting the time and state via `DiffEqBase` to account for `Dual` https://github.com/qojulia/QuantumOptics.jl/blob/e593ae1a51ebc47661322e8ac2c0116085708b2a/src/schroedinger.jl#L123-L124

This would also happed if I would be taking a derivative wrt. time.

Stack trace;
```
ERROR: MethodError: no method matching Float64(::ForwardDiff.Dual{ForwardDiff.Tag{typeof(Ftdop), Float64}, Float64, 1})

Closest candidates are:
  (::Type{T})(::Real, ::RoundingMode) where T<:AbstractFloat
   @ Base rounding.jl:207
  (::Type{T})(::T) where T<:Number
[...]

Issue #357, opened 2023-01-01
Adding ForwardDiff support for all solvers


AmitRotem:
Current fix for supporting Dual number in the Schroedinger solver https://github.com/qojulia/QuantumOptics.jl/pull/356

TODO;
- Implement in other solvers

Related issue;
`Complex{Dual}` support in `DiffEqBase.jl` (see https://github.com/SciML/DiffEqBase.jl/pull/860)

Lightup1:
Happy to hear that the Schroedinger solver is now supporting Dual number and then naturally the Forwarddiff. That's great.

PR #378, opened 2023-12-25
Promote TimeDependentSum to Dual when using ForwardDiff

AmitRotem:
Issue https://github.com/qojulia/QuantumOptics.jl/issues/376

### DIFF
--- a/src/schroedinger.jl
+++ b/src/schroedinger.jl
@@ -58,7 +58,7 @@ function schroedinger_dynamic(tspan, psi0::T, H::AbstractTimeDependentOperator;
     kwargs...) where {B,Bp,T<:Union{AbstractOperator{B,Bp},StateVector{B}}}
     promoted_tspan, psi0 = _promote_time_and_state(psi0, H, tspan)
     if promoted_tspan !== tspan # promote H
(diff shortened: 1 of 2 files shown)


src/schroedinger.jl
```diff
@@ -56,7 +56,13 @@ end
 
 function schroedinger_dynamic(tspan, psi0::T, H::AbstractTimeDependentOperator;
     kwargs...) where {B,Bp,T<:Union{AbstractOperator{B,Bp},StateVector{B}}}
-    schroedinger_dynamic(tspan, psi0, schroedinger_dynamic_function(H); kwargs...)
+    promoted_tspan, psi0 = _promote_time_and_state(psi0, H, tspan)
+    if promoted_tspan !== tspan # promote H
+        promoted_H = TimeDependentSum(H.coefficients, H.static_op.operators; init_time=zero(promote_type(eltype(promoted_tspan), typeof(current_time(H)))))
```
amilsted, line 61: Could even just set `init_time` to `first(promoted_tspan)`?

amilsted:
Thanks @AmitRotem! Looks good. Btw, do we support ForwardDiff for the other evolution methods yet?

AmitRotem:
As far as I know only for schroedinger.
In the past we've opened an issue for that; https://github.com/qojulia/QuantumOptics.jl/issues/357

Issue #386, opened 2024-04-25
usability/documentation improvement: MethodError typehints when picking the wrong master equation function and better `see also` sections in the documentation


Krastanov:
See https://discourse.julialang.org/t/quantumoptics-jl-time-dependent-hamiltonian-cannot-run/113448

When someone picks `master` instead of `master_dynamic`, the `MethodError` message is not particularly clear. We can use the error message hint functionality to tell them they should probably use `master_dynamic`. We should also add references in a `see also` section so that people looking into one of the methods can easily see there is a related method.

gfq960906:
Provide an example to supplement this question：

When we use the time-dependent Hamiltonian
```
function Hami(t,rho)
    detuning = Delta_0*t + (2*delta*sin((omega_0*t)/2)^2)/omega_0
    # detuning = Delta_0*t + (2*delta*sin((omega_0*t)/2)^2)/omega_0
    
    H1 = Omega/2*exp(-1im*(beta*pi/2-alpha-pi/2))*exp(-1im*detuning)*sr⊗dagger(minus) + Omega/2*exp(1im*(beta*pi/2-alpha-pi/2))*exp(1im*detuning)*minus⊗dagger(sr)
    H2 = Omega/2*exp(-1im*(beta*pi/2-alpha-pi/2))*exp(-1im*detuning)*sr⊗dagger(s1) + Omega/2*exp(1im*(beta*pi/2-alpha-pi/2))*exp(1im*detuning)*s1⊗dagger(sr)
 
    return tensor(H1, eye4) + tensor(eye4, H2) + Urr*tensor(sr⊗dagger(sr), sr⊗dagger(sr))
end
```
we need to run the evolution by 
```
tout, pops = timeevolution.schroedinger_dynamic(tlist, psi_0, Hami)
```

a-eghrari:
I would like to work on this as my first issue. From the discussion, I understood that we want to add `See also: master_dynamic` to the documentation of the function `master`. Something like
```
"""
    timeevolution.master(tspan, rho0, L; <keyword arguments>)

Time-evolution according to a master equation with a Liouvillian superoperator `L`.
# Arguments
* `tspan`: Vector specifying the points of time for which output should
        be displayed.
* `rho0`: Initial density operator. Can also be a state vector which is
        automatically converted into a density operator.
* `L`: Superoperator defining the right-hand-side of the master equation.
* `fout=nothing`: If given, this function `fout(t, rho)` is called every time
[...]

PR #388, opened 2024-05-01
Adding hint for master solver function

a-eghrari:
Adding hint for master equation solver when there is a `MethodError`. I also referenced `master_dynamic` in the `see also` section of `master`. [#386](https://github.com/qojulia/QuantumOptics.jl/issues/386)

### DIFF
--- a/src/QuantumOptics.jl
+++ b/src/QuantumOptics.jl
@@ -37,6 +37,7 @@ end
 include("timecorrelations.jl")
 include("spectralanalysis.jl")
 include("semiclassical.jl")
+include("debug.jl")
 module stochastic
     include("stochastic_base.jl")
     include("stochastic_definitions.jl")
--- a/src/debug.jl
+++ b/src/debug.jl
@@ -1,0 +1,12 @@
+function __init__()
+    if isdefined(Base.Experimental, :register_error_hint)
+
+        Base.Experimental.register_error_hint(MethodError) do io, exc, argtypes, kwargs
+            if exc.f == timeevolution.master
+                printstyled(io, "\nHint", color=:green)
+                print(io, ": If your Hamiltonian is time-dependent, then you may want to use master_dynamic instead of master to solve evolution.")
+            end
+        end
+
+    end
+end
--- a/src/master.jl
+++ b/src/master.jl
@@ -115,6 +115,8 @@ Time-evolution according to a master equation with a Liouvillian superoperator `
         permanent! It is still in use by the ode solver and therefore must not
         be changed.
 * `kwargs...`: Further arguments are passed on to the ode solver.
+
+See also: [`master_dynamic`](@ref)
 """
 function master(tspan, rho0::Operator, L::SuperOperator; fout=nothing, kwargs...)
     # Rewrite rho as Ket and L as Operator


Krastanov:
something weird is happening with the tests on x86 architectures, but it does not seem it is due to this pull request. The issue is during the precompilation of SparseArrays. I do not believe it should block this PR as it is unrelated, but we should at least file an issue and see whether there was potentially a accidental break in a non-breaking release of the SparseArrays library

Krastanov:
@a-eghrari , is it possible to add some checks against false positives here.

For instance, if someone has completely nonsensical arguments for `master` the hint will be a bit misleading and should be not shown. Could you make the hint show only if there is an appropriate number of arguments (3 or 4) and the second argument is already correct (of type Operator)?

a-eghrari:
@Krastanov, Thanks for your explanation.

Sure, that is feasible. In addition, I can also type-check the third argument against something like `Union{Function, AbstractTimeDependentOperator}` so we can then be sure that the user is using the wrong function to solve time evolution. What do you think of that?

Upon going through examples and source code, I have noticed that both master and Shrodinger solvers are equipped with a check method (`_check_const`) which throws an `ArgumentError` If the user uses subtypes of `AbstractTimeDependentOperator` for Hamiltonian. Even though I see the method `is_const` is dispatched for the `Function` type as well, it seems it does not work since master and Shrodinger solvers have not been dispatched on `Function`s. For example ([This is example is from QuantumOptics.jl'd documentation](https://docs.qojulia.org/timeevolution/timedependent-problems/)):

```
using QuantumOptics

basis = SpinBasis(1//2)
ψ₀ = spindown(basis)
const sx = sigmax(basis)
function H_pump(t, psi)
  return sin(t)*sx
end

H_pump_tdop = TimeDependentSum(sin=>sx)

const J = [sigmam(basis)]
const Jdagger = [sigmap(basis)]
function f(t,rho)
    H = sin(t)*sx
    return H, J, Jdagger
end

tspan = [0:0.1:10;]

tout, ψₜ = timeevolution.schroedinger_dynamic(tspan, ψ₀, H_pump) 
#Works fine
tout, ψₜ = timeevolution.schroedinger(tspan, ψ₀, H_pump) 
# Throws a MethodError 

tout, ψₜ = timeevolution.schroedinger_dynamic(tspan, ψ₀, H_pump_tdop) 
#Works fine
tout, ψₜ = timeevolution.schroedinger(tspan, ψ₀, H_pump_tdop)
# Throws an ArgumentError with the message: "You are attempting to use a time-dependent dynamics generator (a Hamiltonian or Lindbladian) with a solver that assumes constant dynamics. To avoid errors, please use the _dynamic solvers instead, e.g. schroedinger_dynamic instead of schroedinger"

tout, ρₜ = timeevolution.master_dynamic(tspan, ψ₀, f)
#Works fine
tout, ρₜ = timeevolution.master(tspan, ψ₀, f)
# Throws a MethodError 
tout, ρₜ = timeevolution.master(tspan, ψ₀, H_pump_tdop, J)
# Throws an ArgumentError with the message: "You are attempting to use a time-dependent dynamics generator (a Hamiltonian or Lindbladian) with a solver that assumes constant dynamics. To avoid errors, please use the _dynamic solvers instead, e.g. schroedinger_dynamic instead of schroedinger"
tout, ρₜ = timeevolution.master(tspan, tensor(ψ₀, dagger(ψ₀)), H_pump_tdop, J)
# Throws an ArgumentError with the message: "You are attempting to use a time-dependent dynamics generator (a Hamiltonian or Lindbladian) with a solver that assumes constant dynamics. To avoid errors, please use the _dynamic solvers instead, e.g. schroedinger_dynamic instead of schroedinger"
```

Isn't it better to use the same method for this problem?

a-eghrari:
I made a mistake and closed this PR.

Krastanov:
Usually it is easy to reopen a pull request but it says that you have deleted the originating repository. Feel free to simply open a new pull request if you have the bandwidth to do so.

And you are right, having better, more thorough `_check_const`, which itself sends better error messages, would be good. In some situations though, especially when there is a "missing method dispatch" type of error, we do not ever reach `_check_const` so probably both are necessary.

PR #396, opened 2024-06-12
Inference improvements for timeevolution

amilsted:
This PR does two things:
* Uses `let` blocks around inner function definitions to avoid boxing of closure variables.
* Uses type parameters for function/callable types in some places to encourage specialization.

User-visible changes should be:
* It is possible to use more generic callable types in some places where `Function` types were previously required.
* Parts of MCWF evolution are stabilized and made allocation free, so MCWF evolution should no longer allocate on each call to the integrand.

### DIFF
--- a/src/bloch_redfield_master.jl
+++ b/src/bloch_redfield_master.jl
@@ -175,8 +175,8 @@ end
 
 # Integrate with given fout
 function integrate_br(tspan, dmaster_br, rho,
-                transf_op, inv_transf_op, fout::Function;
-                kwargs...)
+                transf_op, inv_transf_op, fout::F;
+                kwargs...) where {F}
     # Pre-allocate for in-place back-transformation from eigenbasis
     rho_out = copy(transf_op)
     tmp = copy(transf_op)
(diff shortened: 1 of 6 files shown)


Krastanov (approved):
Looks good to me! If you have the bandwidth, I would suggest adding some tests with `@allocated`. Good to merge either way.

```
allocated(f::F) where {F} = @allocated f()
somefunc() = something_with_a_solver...
@test allocated(somefunc) < some_small_number
```

like done here https://github.com/QuantumSavory/QuantumClifford.jl/blob/c1d115d3ee8a929279b724df1cc63fd00a9b255a/test/test_allocations.jl#L6

amilsted:
Think indeed I don't think I have the bandwidth to add alloc tests right now, but agree it would be nice to have. An alternative is to test the integrand function for allocs directly at runtime and warn the user? Might be too annoying though.

amilsted:
I've realized some pieces of this PR (let blocks where types ought to be inferred) might actually be superfluous and a result of [this](https://julialang.slack.com/archives/C7T968HRU/p1718384393196729).

Krastanov:
My vote would be for this to be merged and an issue to be created (maybe a "good first issue") about creating the allocation benchmarks.

I am fine with the superfluous lets.

For reference to future readers, the link from Ashley is to this slack conversation:


image
image

and these two other issues:

https://github.com/JuliaLang/julia/issues/52635
https://github.com/yuyichao/FunctionWrappers.jl/issues/30

PR #401, opened 2024-07-04
Don't tuplify time-dependent ops if types are homogenous

amilsted:
Requires https://github.com/qojulia/QuantumOpticsBase.jl/pull/171 to make allocs test pass.


david-pl:
Looks good to me!

PR #402, opened 2024-07-23
Rates spectrum

ChristophHotter:


test/test_timecorrelations.jl
```diff
@@ -2,7 +2,7 @@ using Test
 using QuantumOptics
 using LinearAlgebra
 
-@testset "timecorrelations" begin
```
Krastanov, line 5: Was removing the `@testset` on purpose? That can mess up a bit how tests get reported in the log.

ChristophHotter: No, sorry. I forgot to uncomment. Thanks for pointing this out, I will fix it!

PR #409, opened 2024-08-09
Implement ForwardDiff in master solvers and general DiffEq problems on QO types

apkille:
- Adds feature discussed in #357 for master solvers.
- I defined `DiffEqBase.promote_u0` methods for QO types and removed `rebuild` because it was no longer needed. Once https://github.com/qojulia/QuantumOpticsBase.jl/pull/172 is merged, we can use ForwardDiff in the standard SciML approach. For example:
```
using OrdinaryDiffEq, ForwardDiff, QuantumOptics

b = SpinBasis(10//1)
psi0 = spindown(b)
H(p) = p[1]*sigmax(b) + p[2]*sigmam(b)
f!(dpsi, psi, p, t) = timeevolution.dschroedinger!(dpsi, H(p), psi)
function cost(p)
    prob = ODEProblem(f!, psi0, (0.0, pi), p)
    sol = solve(prob, DP5(); save_everystep=false)
    return 1 - norm(sol[end])
end

julia> ForwardDiff.gradient(cost, [10.0, -3.0])
2-element Vector{Float64}:
 -151.2972676200132
  -67.02418550025618
```


Project.toml
```diff
@@ -7,6 +7,7 @@ Arpack = "7d9fca2a-8960-54d3-9f78-7d1dccf2cb97"
 DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
 DiffEqCallbacks = "459566f4-90b8-5000-8ac3-15dfb0a30def"
 FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
+FiniteDiff = "6a86dc24-6348-571c-b903-95158fe2bd41"
```
Krastanov, line 10: this should be a test dependency, not a dependency for the project as a whole

apkille: That's my bad, should be good now.

apkille:
@krastanov Our current ForwardDiff tests are a bit finicky because we are testing numerical equality between ForwardDiff and some implementation of finite difference methods. Should we just increase the abstol or would that be too fragile? Or maybe there's a better way to test ForwardDiff here? Maybe we just use `@test_nowarn`?

Krastanov:
increasing the tolerance should be fine (not quite sure whether it is the abs or rel one though)

apkille:
@krastanov is this good to merge?

Krastanov:
The tests currently only test whether operators and their linear algebra works, right? There are no new tests that check the actual solvers? Should we be closing #357 (I guess what tests we need here depends on the answer to that question)?

Do `schroedinger`, `master`, and the others work as well?

Could you add what you have in the first comment (the PR description) to the tests as well?

apkille:
> The tests currently only test whether operators and their linear algebra works, right? There are no new tests that check the actual solvers? Should we be closing #357 (I guess what tests we need here depends on the answer to that question)?

@krastanov the tests do check the master solvers (`timeevolution.master`, `timeevolution.master_h` and `timeevolution.master_nh`), that's what the new testset is doing.

apkille:
> Do `schroedinger`, `master`, and the others work as well?

The `ForwardDiff.Dual` only propagates through `schroedinger`, `schroedinger_dynamic`, and `master` (see #378 as well). I was going to implement ForwardDiff for `master_dynamic` and  `master_nh_dynamic` in this PR, but it's slightly involved because of how the dynamic solvers currently work with input functions that create a tuple of, so I'm saving it for another submission. For instance, `master_dynamic` is written as 
```
timeevolution.master_dynamic(tspan, rho0, f; <keyword arguments>)
``` 
where `f(t, rho) -> (H, J, Jdagger)` or `f(t, rho) -> (H, J, Jdagger, rates)`. So it's tricky propagating the duals here. Same with semiclassical solvers. It's straightforward if we just define the differential equation problem from the start 😃, we have only to write a few lines of code in the internals.

apkille:
> Could you add what you have in the first comment (the PR description) to the tests as well?

Yes, I can do that.

Krastanov:
Wonderful work! Thank you for contributing it!

Issue #406, opened 2024-07-31
Cache input to master derivative functions


apkille:
In several of the master solver functions, e.g., `dmaster_h!`, a cached copy of `drho` is required as input: 
```
dmaster_h!(drho, H, J, Jdagger, rates, rho, drho_cache)
```
Presumably, this is to improve performance of solvers such as `timeevolution.master`. However, this feels excessive if we don't want to use the solver, and instead create an ODE problem directly with our density operators and `dmaster_h!`. For example, it would be great if we could define an in-place method without the cache input, say
```
f!(drho, rho, p, t) = timeevolution.dmaster_h!(drho, H, J, Jdagger, rates, rho)
prob  = ODEProblem(f!, rho0, (t0, t1))
```
This is a natural interface that I'm aiming for with all of the recent broadcasting work (#404, https://github.com/qojulia/QuantumOpticsBase.jl/pull/172). Do the QO maintainers have any comments or oppositions to defining such a method in QO.jl?

david-pl:
It shouldn't create any ambiguities to add that method, so go ahead :+1:

Krastanov:
this might be related https://github.com/qojulia/QuantumOptics.jl/issues/298

apkille:
@krastanov to build off our discussion a workaround is assigning the cached `drho` to the `p` parameter of the ODEProblem object, which prevents us from caching at each time step if we use OrdinaryDiffEq.jl directly. So we don't have to define any new methods. For example, using `dmaster_h!` which is already a part of the public API, we have
```
using QuantumOptics, OrdinaryDiffEq

b = FockBasis(5)
psi0 = fockstate(b, 4)
rho0 = dm(psi0)
tmp = copy(rho0)
H = number(b)
J = [destroy(b), create(b)]
Jdagger = dagger.(J)
rates = [0.7, 0.5]

f!(drho, rho, p, t) = timeevolution.dmaster_h!(drho, H, J, Jdagger, rates, rho, p)
prob = ODEProblem(f!, rho0, (0.0, 1.0), tmp)
sol = solve(prob, DP5(); save_everystep=false)
```

PR #404, opened 2024-07-30
Broadcasting for semiclassical objects

apkille:
@david-pl @Krastanov @amilsted
- Added an interface for broadcasting `State` objects, which enables use of DiffEq solvers
- With these changes, ODE problems for `State`s can be naturally written as the following example:
```
using QuantumOptics, OrdinaryDiffEq, BenchmarkTools

b = SpinBasis(1//2)
psi0 = spindown(b)
u0 = ComplexF64[0.5, 0.75] 
sc = semiclassical.State(psi0, u0)
t₀, t₁ = (0.0, pi)
const σx = sigmax(b)

fquantum(t, q, u) = σx + cos(u[1])*identityoperator(σx)
fclassical!(du, u, q, t) = (du[1] = sin(u[2]); du[2] = 2*u[1])
f!(dstate, state, p, t) = semiclassical.dschroedinger_dynamic!(dstate, fquantum, fclassical!, state, t)
prob = ODEProblem(f!, sc, (t₀, t₁))

julia> @benchmark sol = solve($prob, DP5(); save_everystep=false)
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  50.166 μs … 133.706 ms  ┊ GC (min … max):  0.00% … 99.93%
 Time  (median):     52.583 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   70.621 μs ±   1.339 ms  ┊ GC (mean ± σ):  23.35% ±  3.73%

       ▃█▅                                                      
  ▁▂▃▆████▇▃▂▂▂▂▂▂▂▂▂▃▃▃▃▂▃▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  50.2 μs         Histogram: frequency by time         69.8 μs <

 Memory estimate: 93.81 KiB, allocs estimate: 1136.

julia> @benchmark semiclassical.schroedinger_dynamic([$t₀, $t₁], $sc, $fquantum, $fclassical!)
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  117.875 μs … 137.310 ms  ┊ GC (min … max):  0.00% … 99.87%
 Time  (median):     122.959 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   148.349 μs ±   1.379 ms  ┊ GC (mean ± σ):  15.47% ±  6.25%

[...]

PR #410, opened 2024-08-11
Add derivative methods without caches

apkille:
Addresses #406. This PR makes defining in-place functions straightforward (that is, the user does not have to input caches or adjoints of operators) for any solver. This is important given the recent merges (#404, https://github.com/qojulia/QuantumOpticsBase.jl/pull/172) to have a SciML interface in QO.jl. The following example illustrates the motivation for this PR.

```
b = FockBasis(10)
psi0 = fockstate(b, 4)
rho0 = dm(psi0)
H = number(b)
J = [destroy(b), create(b)]
rates = [0.7, 0.5]
```

Before, we would have to define an in-place function as
```
f!(drho, rho, p, t) = timeevolution.dmaster_h!(drho, H, J, dagger.(J), rates, rho, copy(drho))
```
Now, we can instead simply write
```
f!(drho, rho, p, t) = timeevolution.dmaster_h!(drho, H, J, rates, rho)
```

And solve the ODE as follows:
```
prob = ODEProblem(f!, rho0, (0.0, 1.0))
sol = solve(prob, DP5(); save_everystep=false)
```

### DIFF
--- a/src/master.jl
+++ b/src/master.jl
@@ -280,6 +280,7 @@ end
 
 """
     dmaster_h!(drho, H, J, Jdagger, rates, rho, drho_cache)
+    dmaster_h!(drho, H, J, rates, rho)
 
 Update `drho` according to a master equation given in standard Lindblad form.
 A cached copy `drho_cache` of `drho` is used as a temporary saving step.
@@ -332,10 +333,11 @@ function dmaster_h!(drho, H, J, Jdagger, rates::AbstractMatrix, rho, drho_cache)
     return drho
 end
 
-dmaster_h!(drho, H, J, Jdagger, rates, rho) = dmaster_h!(drho, H, J, Jdagger, rates, rho, copy(drho))
+dmaster_h!(drho, H, J, rates, rho) = dmaster_h!(drho, H, J, dagger.(J), rates, rho, copy(drho))
 
 """
     dmaster_nh!(drho, Hnh, Hnh_dagger, J, Jdagger, rates, rho, drho_cache)
+    dmaster_nh!(drho, Hnh, J, rates, rho)
 
 Updates `drho` according to a master equation given in standard Lindblad form.
 The part of the Liuovillian which can be written as a commutator should be
@@ -375,7 +377,7 @@ function dmaster_nh!(drho, Hnh, Hnh_dagger, J, Jdagger, rates::AbstractMatrix, r
     return drho
 end
 
-dmaster_nh!(drho, Hnh, Hnh_dagger, J, Jdagger, rates, rho) = dmaster_nh!(drho, Hnh, Hnh_dagger, J, Jdagger, rates, rho, copy(drho))
+dmaster_nh!(drho, Hnh, J, rates, rho) = dmaster_nh!(drho, Hnh, dagger(Hnh), J, dagger.(J), rates, rho, copy(drho))
 
 """
     dmaster_liouville!(drho,L,rho)
@@ -393,6 +395,7 @@ end
 
 """
     dmaster_h_dynamic!(drho, f, rates, rho, drho_cache, t)
+    dmaster_h_dynamic!(drho, f, rates, rho, t)
 
 Computes the Hamiltonian and jump operators as `H,J,Jdagger=f(t,rho)` and
 update `drho` according to a master equation. Optionally, rates can also be
@@ -418,6 +421,7 @@ dmaster_h_dynamic!(drho, f::F, rates, rho, t) where {F} = dmaster_h_dynamic!(drh
 
 """
     dmaster_nh_dynamic!(drho, f, rates, rho, drho_cache, t)
+    dmaster_nh_dynamic!(drho, f, rates, rho, t)
 
 Computes the non-hermitian Hamiltonian and jump operators as
 `Hnh,Hnh_dagger,J,Jdagger=f(t,rho)` and update `drho` according to a master
--- a/src/master.jl
+++ b/src/master.jl
@@ -332,6 +332,8 @@ function dmaster_h!(drho, H, J, Jdagger, rates::AbstractMatrix, rho, drho_cache)
     return drho
 end
 
+dmaster_h!(drho, H, J, Jdagger, rates, rho) = dmaster_h!(drho, H, J, Jdagger, rates, rho, copy(drho))
+
 """
     dmaster_nh!(drho, Hnh, Hnh_dagger, J, Jdagger, rates, rho, drho_cache)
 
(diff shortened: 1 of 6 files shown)


apkille:
@krastanov @david-pl pinging for review.

Krastanov:
For the following:

```
f!(drho, rho, p, t) = timeevolution.dmaster_h!(drho, H, J, dagger.(J), rates, rho, copy(drho))
```

does this mean that each step of the solver will be causing an allocation (the `copy`)?

Krastanov:
And related to my question, do we want to create something like this:

```
struct dmaster_h_cache{T,S}
    Jdagger::SomethingSpecific{T}
    rhocache::SomethingElse{S}
end

function (s::dmaster_h_cache)(drho, H, J, rates, rho)
    timeevolution.dmaster_h!(drho, H, J, s.Jdagger, rates, rho, s.rhocache))
end
```

but a bit cleaner and with better names

apkille:
> does this mean that each step of the solver will be causing an allocation (the `copy`)?

@Krastanov yes, that is the case. We actually get a really nice performance speedup here using just the ODE interface. Although there's some allocation overhead, the speed is much better (here I set `b = FockBasis(100)`) and maintains good performance as the problem scales:
```
julia> @benchmark solve($prob, DP5(); save_everystep=false)
BenchmarkTools.Trial: 263 samples with 1 evaluation.
 Range (min … max):  18.140 ms …  22.007 ms  ┊ GC (min … max): 0.00% … 15.85%
 Time  (median):     18.723 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   19.052 ms ± 765.057 μs  ┊ GC (mean ± σ):  3.10% ±  3.71%

    ▃▂█▄▂ ▃                                                     
  ▄▅███████▅▄▃▃▃▅▃▄▅▃▅▆▅▅▅▆▄▅▆▅▆▄▃▃▄▄▄▃▃▃▃▄▃▄▃▃▃▁▁▃▃▁▁▁▃▁▁▁▁▁▃ ▃
  18.1 ms         Histogram: frequency by time         21.5 ms <

 Memory estimate: 18.58 MiB, allocs estimate: 693.

julia> @benchmark timeevolution.master_h([$0.0, $1.0], $rho0, $H, $J)
BenchmarkTools.Trial: 109 samples with 1 evaluation.
 Range (min … max):  45.506 ms …  48.548 ms  ┊ GC (min … max): 0.00% … 5.60%
 Time  (median):     45.900 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   46.005 ms ± 505.631 μs  ┊ GC (mean ± σ):  0.27% ± 1.03%

      ▆▆▆▇▄▄█▆▂                                                 
  ▄▄▄▆█████████▆▃▃▁▁▁▁▁▁▁▁▁▁▁▁▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄▁▁▃▁▁▁▃▁▁▁▁▁▁▁▃ ▃
  45.5 ms         Histogram: frequency by time         48.3 ms <

 Memory estimate: 3.12 MiB, allocs estimate: 131.
```

apkille:
> And related to my question, do we want to create something like this:
> 
> ```
> struct dmaster_h_cache{T,S}
>     Jdagger::SomethingSpecific{T}
>     rhocache::SomethingElse{S}
> end
> 
> function (s::dmaster_h_cache)(drho, H, J, rates, rho)
>     timeevolution.dmaster_h!(drho, H, J, s.Jdagger, rates, rho, s.rhocache))
> end
> ```
> 
> but a bit cleaner and with better names

I think this would be really good to implement. I'd say it's outside the scope of this PR (as here I'm just creating more straightforward methods), but I'd be happy to introduce this to our solvers.

Krastanov:
> > does this mean that each step of the solver will be causing an allocation (the `copy`)?
> 
> @Krastanov yes, that is the case. We actually get a really nice performance speedup here using just the ODE interface. Although there's some allocation overhead, the speed is much better (here I set `b = FockBasis(100)`) and maintains good performance as the problem scales:
> 
> ```
> julia> @benchmark solve($prob, DP5(); save_everystep=false)
> BenchmarkTools.Trial: 263 samples with 1 evaluation.
>  Range (min … max):  18.140 ms …  22.007 ms  ┊ GC (min … max): 0.00% … 15.85%
>  Time  (median):     18.723 ms               ┊ GC (median):    0.00%
>  Time  (mean ± σ):   19.052 ms ± 765.057 μs  ┊ GC (mean ± σ):  3.10% ±  3.71%
> 
>     ▃▂█▄▂ ▃                                                     
>   ▄▅███████▅▄▃▃▃▅▃▄▅▃▅▆▅▅▅▆▄▅▆▅▆▄▃▃▄▄▄▃▃▃▃▄▃▄▃▃▃▁▁▃▃▁▁▁▃▁▁▁▁▁▃ ▃
>   18.1 ms         Histogram: frequency by time         21.5 ms <
> 
>  Memory estimate: 18.58 MiB, allocs estimate: 693.
> 
> julia> @benchmark timeevolution.master_h([$0.0, $1.0], $rho0, $H, $J)
> BenchmarkTools.Trial: 109 samples with 1 evaluation.
>  Range (min … max):  45.506 ms …  48.548 ms  ┊ GC (min … max): 0.00% … 5.60%
>  Time  (median):     45.900 ms               ┊ GC (median):    0.00%
>  Time  (mean ± σ):   46.005 ms ± 505.631 μs  ┊ GC (mean ± σ):  0.27% ± 1.03%
> 
>       ▆▆▆▇▄▄█▆▂                                                 
>   ▄▄▄▆█████████▆▃▃▁▁▁▁▁▁▁▁▁▁▁▁▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄▁▁▃▁▁▁▃▁▁▁▁▁▁▁▃ ▃
>   45.5 ms         Histogram: frequency by time         48.3 ms <
> 
>  Memory estimate: 3.12 MiB, allocs estimate: 131.
> ```

This confuses me a lot. What is going wrong in `master_h` that this new method which should be strictly slower (because of all the extra copies) is actually faster?

Is there a potential future version of `timeevolution.dmaster_h!(drho, H, J, rates, rho)` that does not make copies on each step? I feel reluctant including a new approach that is slower than the currently existing approach, even if it is more convenient. Can we make this into something that is as fast as the already existing method?

apkille:
@krastanov `time evolution.master_h` calls `dmaster_h!` in its backend. Before I submitted the PR,`dmaster_h!` already copied `drho` at each step. Here, I am simply defining a method that makes this function usable, that is, not providing an argument that copies `drho`:
```
dmaster_h!(drho, H, J, rates, rho) = dmaster_h!(drho, H, J, dagger.(J), rates, rho, copy(drho))
```
Maybe I am thinking about this the wrong way but this was my thought process.

Issue #407, opened 2024-08-01
Update the benchmark suite and bring it into the CI runner [$400]


Krastanov:
<details>
<summary><strong>Bug bounty logistic details</strong> (click to expand)</summary>

To claim exclusive time to work on this bounty either post a comment here or message [skrastanov@umass.edu](mailto:skrastanov@umass.edu) with:

- your name
- github username
- **(optional)** a brief list of previous pertinent projects you have engaged in

Currently the project is claimed by `no one` until `...`.

If you want to, you can work on this project without making a claim, however claims are encouraged to give you and other contributors peace of mind. Whoever has made a claim takes precedence when solutions are considered.

You can always propose your own funded project, if you would like to contribute something of value that is not yet covered by an official bounty.
</details>

# Update the benchmark suite and bring it into the CI runner [$400]

QuantumOptics.jl has an in-depth benchmarking suite that has bitrotted in the last few years. To claim this bounty:

- Set up a benchmark-on-CI infrastructure (e.g. similar to the one in QuantumClifford.jl) that runs QuantumOptics-specific benchmarks on each pull request
- Include all old benchmarks in that infrastructure (reusing the existing benchmark suite as much as possible) and make it easy to add new micro-benchmarks as new features get developed.

The above are for the sake of the developers of this package. However, this bounty also requires the following to be done for the sake of users:

- Fix up the capability to run benchmarks in comparison to popular python packages (qutip) and add comparative Julia packages (QuantumToolbox.jl)
- Create a makefile that let's a developer to easily update the webpage presenting these comparative benchmarks to the user.

**Required skills**: Understanding of Julia's and Python's benchmarking infrastructure.

**Reviewer**: Stefan Krastanov and/or @david-pl 

**Duration**: 2 months

#### Payout procedure:

The Funding for these bounties comes from the National Science Foundation and from the NSF Center for Quantum Networks. The payouts are managed by the NumFOCUS foundation and processed in bulk once every two months. If you live in a country in which NumFOCUS can make payments, you can participate in this bounty program.

[Click here for more details about the bug bounty program.](https://github.com/QuantumSavory/.github/blob/main/BUG_BOUNTIES.md)

nenadilic84:
I'd like to claim this bounty.

[...]

PR #411, opened 2024-08-11
Implement benchmarks for broadcasting QO types

apkille:
Integrates very simple benchmarking into CI, specifically for bringing QO.jl up to speed with the SciML interface ([#404](https://github.com/qojulia/QuantumOptics.jl/issues/298)). Note, at this current moment, this is **not** me reserving the following bounty #407. I am merely submitting this PR to benchmark work in merged PRs (#404, https://github.com/qojulia/QuantumOpticsBase.jl/pull/172). There is still a substantial amount of work to be done in migrating other QO benchmarks to this suite.

I should also mention: to add support for solving  SDEs with QO types, I defined the method `wiener_randn!` from DiffEqNoiseProcess.jl, which meant adding it as a dependency (this is a requirement for custom array types https://github.com/SciML/StochasticDiffEq.jl/pull/579#event-13828699414).


apkille:
Ah, StochasticDiffEqs hasn't released a new version yet with my changes

apkille:
Converting to draft until SDE.jl bumps version, which should be soon as discussed on slack.

apkille:
@Krastanov pinging for review. The performance tracking test is failing because the benchmark result of this branch cannot be compared with master. It should work for future PRs if we merge this. Correct me if I'm wrong.

src/stochastic_base.jl
```diff
@@ -104,3 +105,5 @@ function _check_const(op)
     end
     nothing
 end
+
+DiffEqNoiseProcess.wiener_randn!(rng::AbstractRNG,rand_vec::T) where {T<:Union{Bra,Ket,Operator}} = randn!(rng,rand_vec.data)
```
Krastanov, line 109: Is this a bugfix for something?

benchmark/benchmarks.jl
```diff
@@ -0,0 +1,113 @@
...
+    master!(drho, rho, p, t) = timeevolution.dmaster_h!(drho, Hobj, [Jobj], [Jdag], rates, rho, copy(obj))
+    stoch_master!(drho, rho, p, t) = timeevolution.dmaster_h!(drho, Hsobj, [Jobj], [Jdag], rates, rho, copy(obj))
+    prob = SDEProblem(master!, stoch_master!, obj, (t₀, t₁))
+end
+
+for dim in (1//2, 20//1, 50//1, 100//1)
+    for prob in zip(("schroedinger", "master"), (:(bench_schroedinger), :(bench_master)))
+        name, bench = (prob[1], prob[2])
+        # benchmark solving ODE problems on data of QO types
+        SUITE[name]["pure"][string(dim)] = @benchmarkable solve(eval($bench)($dim; pure=true), DP5(); save_everystep=false)
```
Krastanov, line 102: I am a bit worried that `eval` is part of the benchmark. Is there any change if instead you do `solve(prob) setup=(prob=eval(...))`?

Project.toml
```diff
@@ -34,7 +35,7 @@ OrdinaryDiffEq = "5, 6"
 QuantumOpticsBase = "0.3, 0.4, 0.5"
 RecursiveArrayTools = "2, 3"
 Reexport = "0.2, 1.0"
-StochasticDiffEq = "6"
+StochasticDiffEq = "6, 6.68.0"
```
Krastanov, line 38: this should probably just be one single entry `6.68` if you want to refuse `6.67`

Project.toml
```diff
@@ -6,6 +6,7 @@ version = "1.1.1"
 Arpack = "7d9fca2a-8960-54d3-9f78-7d1dccf2cb97"
 DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
 DiffEqCallbacks = "459566f4-90b8-5000-8ac3-15dfb0a30def"
+DiffEqNoiseProcess = "77a26b50-5914-5dd7-bc55-306e6241c503"
```
Krastanov, line 9: this requires a compat bound (I am surprised Aqua did not catch this... oh, we do not have Aqua in QuantumOptics.jl...)

benchmark/benchmarks.jl
```diff
@@ -0,0 +1,113 @@
...
+using LinearAlgebra
+using PkgBenchmark
+
+
+const SUITE = BenchmarkGroup()
+
+prob_list = ("schroedinger", "master", "stochastic_schroedinger", "stochastic_master")
+for prob in prob_list
+    SUITE[prob] = BenchmarkGroup([prob])
+    for type in ("pure", "custom")
```
Krastanov, line 14: May we use "QO types" and "Base array types" instead of pure and custom? These confused me for a second

Krastanov (changes requested):
Sorry for the slow review. This is very valuable work, thank you! Do you mind making the following small changes?

src/stochastic_base.jl
apkille: Yes, I had to define this method to enable SDEs to be solved on QO types. I didn't see an easy PR fix to submit to SciML so I defined it here. If this is too sketchy to you, let me know and I can do some more digging to work around this problem.

Project.toml
apkille: 😅 I'll submit an issue...

apkille:
@krastanov pinging for review.

Krastanov:
Thank you! Great to finally have benchmarks as part of CI.

Issue #442, opened 2025-05-19
Static analysis tests for code quality (with JET and Aqua)


Krastanov:
JET and Aqua are two very common tools for static analysis in the Julia ecosystem. They can frequently detect issues and bugs before anyone encounters them. Many other packages in the QuantumOptics and QuantumSavory ecosystem already employ them in their test runners. Below are a few examples:

- QuantumOpticsBase: [aqua](https://github.com/qojulia/QuantumOpticsBase.jl/blob/master/test/test_aqua.jl), [jet](https://github.com/qojulia/QuantumOpticsBase.jl/blob/master/test/test_jet.jl) 
- QuantumClifford: [aqua](https://github.com/QuantumSavory/QuantumClifford.jl/blob/master/test/test_aqua.jl), [jet](https://github.com/QuantumSavory/QuantumClifford.jl/blob/master/test/test_jet.jl)

Notice how in the QuantumOpticsBase the `Aqua` tests have some of them marked as broken, and the JET tests have a number of detected issues marked as not currently solved. Use this as a reference on how to deal with false positives.

### This bounty is about the addition of Aqua and JET checks to QuantumOptics.jl

To complete the bounty you need to:

- submit a PR which enables Aqua checks and fixes the uncovered issues
- submit a PR which enables JET checks. It is not necessary to fix the uncovered issues, as dealing with false positives is difficult, but you need to at least mark the tests as broken
- both PRs need to be approved and merged for the bounty to be awarded

You are strongly encouraged to submit two separate PRs to make reviewing easier and faster. Consult the existing examples of such tests in QuantumOpticsBase.

Other useful resources:

- the docs for [Aqua](https://juliatesting.github.io/Aqua.jl/dev/)
- the docs for [JET](https://aviatesk.github.io/JET.jl/dev/)

If you are new to julia, make sure to:
- install julia using `juliaup` for version management
- look through the [julia manual](https://docs.julialang.org/en/v1/) and [modern workflows](https://modernjuliaworkflows.org/)
- try out VSCode with the [Julia plugin](https://www.julia-vscode.org/) for good IDE support

ErikQQY:
Yeah, sure, thanks for the patient guidance!

PR #446, opened 2025-05-28
Add JET tests

ErikQQY:
Part of #442


ErikQQY:
From local testing, it seems the current changes are sufficient to pass JET analysis.

```julia
julia> JET.test_package(QuantumOptics; target_defined_modules=true)
Test Passed
```

Krastanov (changes requested):
whoops, you added the Manifest.toml to the commit (I guess we never bothered adding it to `.gitignore`). Could you please remove it?

PR #447, opened 2025-05-28
Add Aqua tests

ErikQQY:
Fix: #442

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -24,25 +24,32 @@ StochasticDiffEq = "789caeaf-c7a9-5a7d-9973-96adeb23e2a0"
 WignerSymbols = "9f57e263-0b3d-5e2e-b1be-24f2bb48858b"
 
 [compat]
+Aqua = "0.8.11"
 Arpack = "0.5.1 - 0.5.3"
 DiffEqBase = "6.162"
 DiffEqCallbacks = "2, 3, 4"
 DiffEqNoiseProcess = "5.23.0"
 FFTW = "1"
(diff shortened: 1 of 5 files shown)


Krastanov (changes requested):
Thank you! Below are a few suggestions:

- could you rename the file to `test_aqua.jl` to make it clearer what it is about.
- could you make the piracies explicit (e.g. see how it is done in https://github.com/qojulia/QuantumOpticsBase.jl/blob/master/test/test_aqua.jl ) -- we do not want to disable tests unless we know for a fact that the failure is ok
- same about the `unbound_args` -- what are the actual errors related to those?

Krastanov:
The failures are probably because the file was not explicitly included in `runtests.jl`

ErikQQY:
The main things Aqua can't handle are extending functions from QuantumOpticsBase in phasespace, these functions are originally just simple signatures in QuantumOpticsBase, defining any other dispatch would be considered as type piracy. I think we can have two solutions here:

1. Move these functions from QuantumOpticsBase to QuantumOptics
2. Explicitly tell Aqua these are fine.

Which one do you prefer? @Krastanov

Krastanov:
do you mind addressing the conflicts from the merged JET branch?

Krastanov:
Wonderful! Thanks for the contribution! We also have bounties beyond the UnitaryHACK you can consult here https://github.com/QuantumSavory/.github/blob/main/BUG_BOUNTIES.md

Issue #443, opened 2025-05-19
Switch tests to use to `TestItems.jl` for easier integration with IDEs and other advanced tests runners


Krastanov:
Like any modern language, Julia supports running unit tests as part of the dev workflow. In this package they are in the `test` folder and can be executed by going to package mode and running `] test QuantumOptics`. They are build with the `Test` standard library.

There are testing harness libraries that have been build on top of it, providing many additional features. `TestItems` in particular makes it much easier to run tests in parallel, it integrates with various IDEs, and provides other nice features.

### This bounty is about packaging all of the tests in QuantumOptics and QuantumOpticsBase as `TestItems` tests.

The conversion is fairly straightforward:

For each test file, wrap it in `@testitem "description" ... end`. It would be best to **not** change any other indentation in that file, so that the diffs shown in the eventual PR are simple and easy to review.

Modify the `runtests.jl` file to be much simpler than it currently is, and to use the `TestItems` runner.

The Project.toml file will need minor modifications to add the test dependencies on `TestItems`

You can consult other libraries in this same ecosystem to see how straightforward the change is, e.g. https://github.com/QuantumSavory/QuantumClifford.jl/tree/master/test

To complete this bounty:
- A PR for QuantumOptics that performs the aforementioned conversion and has all tests passing and no lost tests
- A PR for QuantumOpticsBase that performs the aforementioned conversion and has all tests passing and no lost tests
- there might be some minor back-and-forth on editing ENV variables to make sure that tests are filtered properly by the CI runners (see how it is done in QuantumClifford.jl)
- both PRs need to be approved and merged for the bounty to be awarded

If you are new to julia, make sure to:
- install julia using `juliaup` for version management
- look through the [julia manual](https://docs.julialang.org/en/v1/) and [modern workflows](https://modernjuliaworkflows.org/)
- try out VSCode with the [Julia plugin](https://www.julia-vscode.org/) for good IDE support

luffy-orf:
@Krastanov  I noticed that the issue doesn’t mention the reward amount for completing the conversion to TestItems.jl. Could you please clarify what the bounty is for this task?

Krastanov:
This is in preparation for the upcoming Unitary Foundation hackathon. The bounties will be officially announced there later this week, and only at that point will have specific reward amounts.

Krastanov:
@ErikQQY , could you comment here so that github will permit me to assign the issue to you?

ErikQQY:
[...]

PR #448, opened 2025-05-29
Switch to TestItems

ErikQQY:
Fix: #443


Krastanov:
I am not quite sure tests are actually running here. E.g. project coverage used to be above 80%, now it is 15% according to the codecov comment above.

I would suggest:
- checking the runtests.jl of libraries like QuantumClifford and using something in that style
- not removing the Test imports
- wrapping the entire file (imports included) in a testitem instead of converting each testset into a testitem (testitems can contain nested testsets)

Also, FYI this bounty is for QuantumOptics and QuantumOpticsBase both being done

Krastanov (changes requested):
just adding a formal review comment to mark it as "request changes"

test/runtests.jl
```diff
@@ -1,51 +1,5 @@
...
-if length(unavailable_tests) != 0
-    error("The following tests could not be found:\n", join(unavailable_tests, "\n"))
-end
-
-for name=names
-    if startswith(name, "test_") && endswith(name, ".jl")
-        include(name)
-    end
-end
+testfilter = ti->(contains(ti.filename, "test_") && endswith(ti.filename, ".jl") )
```
Krastanov, line 4: it does not seem like you need this filter

test/test_ForwardDiff.jl
```diff
@@ -1,5 +1,6 @@
+@testitem "Test ForwardDiff.jl with QuantumOptics.jl" begin
 using Test
-using OrdinaryDiffEqCore, QuantumOptics
+using OrdinaryDiffEqLowOrderRK, QuantumOptics
```
Krastanov, line 3: any reason for this change?

test/ForwardDiff_long_test.jl
```diff
@@ -1,4 +1,3 @@
-using Test
```
Krastanov, line 1: do not remove these

test/ForwardDiff_long_test.jl
```diff
@@ -47,7 +46,7 @@ function test_vs_fin_diff(fun, p; ε=√eps(eltype(p)), kwargs...)
     isapprox(FD_grad, fin_diff_grad; kwargs...) ? true : (@show abs_diff, rel_diff; false)
 end
 
-@testset "ForwardDiff with schroedinger" begin
+@testitem "ForwardDiff with schroedinger" begin
```
Krastanov, line 49: as long as this stays as a testset it will not be executed.

or even better, mark it as "tags = [:slow]" and add a filter that excludes slow tests

test/test_aqua.jl
```diff
@@ -1,8 +1,7 @@
-using Test
-using QuantumOptics
-
-@testset "Quality Assurance" begin
+@testitem "Quality Assurance" tags = [:aqua] begin
```
Krastanov, line 1: In my initial comment, I mentioned just wrapping the existing testsets in a testitem. What you have done here is not exactly the same.

It would be easier to just NOT modify any of the content of the files except to add `@testitem ... begin` at the very start of the file, and `end` at the end of the file, and to not make any other changes (including not changing indentation). It would also lead to way fewer changes to the files as a whole (and should avoid the need to redefine some variables as globals)

Do not remove various `using` statements either, just keep them below the `@testitem`

Krastanov (changes requested):
This is getting close, thanks! See comments below.

test/test_ForwardDiff.jl
ErikQQY: I believe `DP5` is in OrdinaryDiffEqLowOrderRK.

test/test_aqua.jl
ErikQQY: Sure, I will just add `@testitem`

test/ForwardDiff_long_test.jl
ErikQQY: Oh, should we also include this test file in the testing?

Krastanov: TestItems automatically detects all `@testitem` in the codebase, no need for any explicit inclusions. This particular file was never executed because it is a very slow test -- we just kept it in case of debugging needs, etc. With testitems this can be done more neatly -- mark it as `:slow` and filter it out.

test/test_ForwardDiff.jl
Krastanov: But the test used to pass fine in the past. Why is the change necessary now?

Krastanov:
once the tests pass, this should be good to merge

PR #452, opened 2025-06-11
Fix #432

### LINKED ISSUE #432: We rely on `promote_dual`, removed from upstream DiffEqBase
This line
https://github.com/qojulia/QuantumOptics.jl/blob/0085f66c9f80381e3f8e8a4e3f013b7cea32090c/src/timeevolution_base.jl#L136
relies on a function that is no longer exported by DiffEqBase: https://github.com/SciML/DiffEqBase.jl/pull/1111/files

Test suite now fails with latest DiffEqBase.

apkille:
#432 This fix properly uses `DiffEqBase.anyeltypedual` and `DiffEqBase.isdualtype` rather than `DiffEqBase.promote_dual` defined in `ext/DiffEqBaseForwardDiffExt.jl`. It also removes ForwardDiff.jl as a direct dependency.

Pinging @Krastanov or @amilsted for review.


Krastanov:
Thank you!

PR #455, opened 2025-06-24
Add automatic differentiation test suite and add ForwardDiff support for dynamic master solvers

apkille:
I added a test suite that exhaustively checks automatic differentiation capabilities for each solver. With this PR, FiniteDiff.jl and ForwardDiff.jl are fully supported for schroedinger and master solvers (including their dynamic versions). I added DifferentiationInterface.jl as a test dependency to quickly test other autodiff libraries supported in Julia in the future (such as Zygote.jl and Enzyme.jl). In the future I will also add support for the stochastic, semi-classical, and Monte Carlo solvers.

One note: here I am simply testing whether or not each differentiation operation runs on each solver with random test cases. I'm open to testing for correctness within some numerical tolerance, but I'd imagine we'd have to be extremely careful that the hundreds of tests pass every time, particularly when we have a handful of autodiff libraries supported in the suite.


Krastanov:
I will mark this as a draft, just to organize my review queue a bit. Feel free to mark it back at any time.

